PR #10, opened 2017-06-28
gentoo-ize python

Alessandro-Barbieri:
Python3 is not officially supported in ROS

### DIFF
--- a/ebuild_generator/ebuild.py
+++ b/ebuild_generator/ebuild.py
@@ -235,7 +235,7 @@ class Ebuild(object):
         ret += "\"\n\n"
 
         ret += "REQUIRED_USE=\"${PYTHON_REQUIRED_USE}\"\n\n"
-        
+
         # SLOT
         ret += "SLOT=\"0\"\n"
         # CMAKE_BUILD_TYPE
--- a/ebuild_generator/ebuild.py
+++ b/ebuild_generator/ebuild.py
@@ -147,10 +147,11 @@ class Ebuild(object):
 
         # EAPI=<eapi>
         ret += "EAPI=" + self.eapi + "\n\n"
+        ret += "PYTHON_COMPAT=( python2_7 )\n\n"
         """
         @todo: don't hard code this
         """
-        ret += "inherit cmake-utils eutils\n\n"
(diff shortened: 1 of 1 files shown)


allenh1:
Doesn't this conflict with #12?

Alessandro-Barbieri:
nono, from source builds fine

allenh1:
@Alessandro-Barbieri Conflicts here as well.

Alessandro-Barbieri:
Not all ROS packages will work with python3
see https://github.com/ros/geometry/issues/148 https://github.com/ros/bond_core/issues/23 https://github.com/ros/ros/issues/148
first of all we need a list of package <-> python compatibility
if we want use multiple pythons implementations, reusing (and adapting) the ros-catkin.eclass is much faster as it's already all done

allenh1:
While it's certainly true that not all ROS packages work with Python 3, the vast majority of them do, including all packages needed for desktop_full.

The `ros-python-[distro]` executable needs to be reworked so that this change will be valid as well, since that has the python version hard-coded into it as well (this could probably be passed in as a macro, I just haven't had the time to adjust it myself).

Alessandro-Barbieri:
Installing from source lunar's desktop_full with python3 led to some errors in at least four packages that maybe you didn't noticed because them are wrongly set to nonfatal

allenh1:
Can you please list these packages for me, with their respective errors?

Alessandro-Barbieri:
>Can you please list these packages for me, with their respective errors?

three comments above

tfoote:
It is true that we only officially support python2 however we know of several users who use python3 for various reasons and we do our best to increase compatibility over time as people file tickets. Can we support having 2.7 as the default and if people override it they can select a python3 version and know that they might not be able to use some functionality?

allenh1:
@tfoote that sounds really reasonable to me.

@Alessandro-Barbieri Can you modify the diff to reflect this behavior?

allenh1:
@Alessandro-Barbieri This behavior is implemented in #28, so I'm going to close this PR. I welcome your feedback for the PR above (sorry to close this one -- I just really doubt you'd want to rebase it on top of the recent changes).

PR #11, opened 2017-06-28
do not enforce optimizations

Alessandro-Barbieri:


allenh1:
Thanks for the patch!

The code in place currently is a short-term fix to a bigger problem: OpenCV, in the current ROS version, does not compile if you have architecture flags on.

Do you think you could change this to use the user CFlags without the machine flags?

Alessandro-Barbieri:
do you mean `-march` and `-mtune`?

allenh1:
Yeah, those. Super obnoxious that it breaks things, but it does.

Alessandro-Barbieri:
totally unsure about this, can I have a sample of a build failure?

ebuild_generator/ebuild.py
```diff
@@ -257,6 +261,9 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
 
         # source configuration
         ret += "src_configure() {\n"
+        if self.name == 'opencv3':
+            ret += "    replace-cpu-flags '*' ''\n"
+            ret += "    filter-flags -march=\n"
```
allenh1: I don't think there should be an '=' there.

allenh1:
Other than that, this looks good. Much better than what I had.

ebuild_generator/ebuild.py
```diff
@@ -260,6 +264,8 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
 
         # source configuration
         ret += "src_configure() {\n"
+        if self.name == 'opencv3':
+            ret += "    filter-flags \'-march=*\' \'-mtune=*\'\n"
```
allenh1: I don't think you need to escape the single quotes, since they're in a double quote string. Also, can you verify that this needs '='?

Alessandro-Barbieri: It's taken from
https://www.mail-archive.com/gentoo-alt@lists.gentoo.org/msg07234.html

PR #13, opened 2017-06-28
use arch independent path

Alessandro-Barbieri:


Alessandro-Barbieri:
lib64 is obviously a 64 bits architecture only folder

allenh1:
@Alessandro-Barbieri That's a fair point.

allenh1:
@Alessandro-Barbieri This is good to merge after that conflict is resolved.

allenh1:
Also pulled in with #28

PR #14, opened 2017-06-28
remove spaces in license

Alessandro-Barbieri:


ebuild_generator/ebuild.py
```diff
@@ -176,7 +176,7 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
                 ret += ')"\n'
             else:
                 ret += "LICENSE=\""
-                ret += get_license(self.upstream_license) + "\"\n\n"
+                ret += get_license(self.upstream_license.replace(' ', '')) + "\"\n\n"
```
allenh1: I'd prefer the get_license function have this kind of line for the default return case (the one selected when none of the regexes match).

allenh1:
This brings to light the need for better licenses, though. There are a lot of silly ones in the package.xml files like "ToDo" that are obviously not real licenses. Any ideas as to how this should be dealt with?

ebuild_generator/ebuild.py
Alessandro-Barbieri: where I should remove spaces then?

allenh1: https://github.com/allenh1/ebuild-generator/blob/2709852666fdec9ce54fd904c559fea5e7e83795/ebuild_generator/ebuild.py#L79

allenh1:
Multiple licenses apply. It's a legal requirement of the license to distribute all licenses with it, so it's entirely possible to have a single file under multiple licenses, then the rest of a repo under a different one. It's kinda strange, but so it is.

Alessandro-Barbieri:
Ok then I'll remove `||`

allenh1:
Sounds like a good idea to me.

PR #15, opened 2017-06-28
avoid version duplication in name

Alessandro-Barbieri:


allenh1:
Haha, I was wondering what caused that. Thanks!

PR #16, opened 2017-06-28
removed logical or

Alessandro-Barbieri:
all licenses applies


allenh1:
This looks fine, but it would be useful to have a document that states that this syntax is valid.

https://devmanual.gentoo.org/general-concepts/licenses/

That's the only item I could find, but, unfortunately, it doesn't talk about the case of multiple licenses applying (though it does state the syntax for when at least one of many licenses can apply).

Alessandro-Barbieri:
https://devmanual.gentoo.org/ebuild-writing/variables/index.html#license

Issue #9, opened 2017-06-26
Installation Instructions


allenh1:
The install instructions/usage instructions should be at least minimally present somewhere in this repo.

connect to #26

PR #26, opened 2017-07-01
Rename Project & Move to Upstream Eclass

allenh1:
wait for rebase on top of #25


allenh1:
connects to #9

allenh1:
@tfoote I think that's really reasonable. We'll definitely revisit that in the future.

PR #28, opened 2017-07-03
Fix python

allenh1:
Removes the hard coded python versions entirely!

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -157,13 +157,11 @@ class Ebuild(object):
         ret += " license\n\n"
 
         # EAPI=<eapi>
-        ret += "EAPI=" + self.eapi + "\n\n"
-        """
-        @todo: don't hard code this
-        """
+        ret += "EAPI=" + self.eapi + "\n"
+        ret += "PYTHON_COMPAT=( python{2_7,3_5} )\n\n"
 
         # inherits
-        ret += "inherit ros-cmake\n"
+        ret += "inherit ros-cmake\n\n"
 
         # description, homepage, src_uri
         py_ver = sys.version_info
@@ -222,10 +220,6 @@ class Ebuild(object):
             first = False
 
         ret += "\"\n"
-        """
-        @todo: uh... probably shouldn't force PYTHON 3.5
-        """
-        ret += "PYTHON_DEPEND=\"3::3.5\"\n\n"
         # RDEPEND
         ret += "RDEPEND=\"\n"
         for rdep in sorted(self.rdepends):
@@ -250,9 +244,8 @@ class Ebuild(object):
         ret += "\"\n\n"
 
         # SLOT
-        ret += "SLOT=\"{}\"\n".format(self.distro)
+        ret += "SLOT=\"0\"\n"
         # CMAKE_BUILD_TYPE
-        ret += "CMAKE_BUILD_TYPE=RelWithDebInfo\n"
         ret += "ROS_DISTRO=\"{0}\"\n".format(self.distro)
         ret += "ROS_PREFIX=\"opt/ros/${ROS_DISTRO}\"\n\n"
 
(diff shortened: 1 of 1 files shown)


superflore/generators/ebuild/ebuild.py
```diff
@@ -263,34 +256,32 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
...
             elif self.name == 'catkin':
                 ret += catkin_cmake_args
-            ret += "    cmake-utils_src_configure\n"
+            ret += "    python_foreach_impl ros-cmake_src_configure_internal\n"
             ret += "}\n\n"
 
         if self.name == 'catkin':
             ret += "src_compile() {\n"
-            ret += "    ${CC} ${FILESDIR}/ros-python.c"
+            ret += "    gcc ${FILESDIR}/ros-python.c"
```
allenh1: @Alessandro-Barbieri For some reason, `${CC}` was not working, so this has been reverted.

allenh1:
@Alessandro-Barbieri this should remove the Python dependencies. Does this look acceptable to you?

superflore/generators/ebuild/ebuild.py
aballier: you need to inherit toolchain-funcs and use $(tc-getCC)

superflore/generators/ebuild/ebuild.py
```diff
@@ -263,34 +256,32 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
...
         special_pkgs = ['catkin', 'opencv3', 'stage']
-        catkin_cmake_args = "    local mycmakeargs=(\n"\
-            + "        -DCMAKE_INSTALL_PREFIX=${D%/}${ROS_PREFIX}\n"\
-            + "        -DCMAKE_PREFIX_PATH=${ROS_PREFIX}\n"\
-            + "        -DPYTHON_INSTALL_DIR=lib64/python3.5/site-packages\n"\
+        catkin_cmake_args = "local sitedir=\"$(python_get_sitedir)\"\n"\
+            + "    local mycmakeargs=(\n"\
+            + "        -DCMAKE_INSTALL_PREFIX=/${ROS_PREFIX}\n"\
+            + "        -DCMAKE_PREFIX_PATH=/${ROS_PREFIX}\n"\
+            + "        -DPYTHON_INSTALL_DIR=${sitedir#${EPREFIX}}/${ROS_PREFIX}\n"\
```
Alessandro-Barbieri: Sure that is not `PYTHON_INSTALL_DIR=${ROS_PREFIX}${sitedir#${EPREFIX}}`?

I'm experiencing strange things with this, not all packages respect this variable (roslz4)
For example with roslz4:

it install pythons (`.py` and company) in `/opt/ros/usr/lib64/python2.7/site-packages/`
but it install libraries (`.so`) in `/usr/lib64/python2.7/site-packages/`

if I prefix `/opt/ros/` it became absurdly

pythons (`.py` and company) in `/opt/ros/opt/ros/usr/lib64/python2.7/site-packages/`
libraries (`.so`) in `/opt/ros/usr/lib64/python2.7/site-packages/`

@aballier what is the correct way to install pythons in an offset directory?
@aballier I'm trying to get the `ros-catkin.eclass` working with gentoo's ros but I'm stuck with this issue, where we can discuss further?

superflore/generators/ebuild/ebuild.py
```diff
@@ -157,13 +157,11 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
         ret += " license\n\n"
 
         # EAPI=<eapi>
-        ret += "EAPI=" + self.eapi + "\n\n"
-        """
-        @todo: don't hard code this
-        """
+        ret += "EAPI=" + self.eapi + "\n"
+        ret += "PYTHON_COMPAT=( python{2_7,3_5} )\n\n"
```
Alessandro-Barbieri: Python3 will not work with all packages, so don't blindly put it everywhere. We need a list of the packages known working with python3.

Alessandro-Barbieri:
Why not move all in the eclass?

superflore/generators/ebuild/ebuild.py
allenh1: Thank you! Will fix that.

superflore/generators/ebuild/ebuild.py
allenh1: @Alessandro-Barbieri Looks like I forgot to push the last commits. One moment.

superflore/generators/ebuild/ebuild.py
allenh1: @aballier actually, this isn't needed anymore (after setting up the python paths correctly).

superflore/generators/ebuild/ebuild.py
allenh1: The correct value of `PYTHON_INSTALL_DIR` is now handled in the eclass.

superflore/generators/ebuild/ebuild.py
allenh1: As it is now, there is no such list available (and, due to the sheer number of packages, such a list is likely to be taken on a case-by-case basis). So, I will remove the python3 flag only for those packages which are known to fail. Although Python 3 does not *work* with all packages, it will still install all packages (as exhibited by my ability to install ros-lunar with 3.5).

We'll simply add a disclaimer that Python 3 is not guaranteed to work for all packages.

allenh1: Other than this, are there any other changes that should be made here?

Alessandro-Barbieri: Try pypy and pypy3

Alessandro-Barbieri: > it will still install all packages (as exhibited by my ability to install ros-lunar with 3.5)

If you count silent failures then you "installed" it, the thing that bother me more is that we will never see if a package is building fine or not because the installation process will continue to run in case of errors. This needs to be fixed upstream.

allenh1: Maybe, but that should be handled on a case-by-base basis, as I don't have a list available. Do you see any reason that this should not be merged, aside from this? This is not meant to be a permanent fix, but rather another step along the journey.

superflore/generators/ebuild/ebuild.py
```diff
@@ -250,9 +244,10 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
         ret += "\"\n\n"
 
         # SLOT
-        ret += "SLOT=\"{}\"\n".format(self.distro)
+        ret += "SLOT=\"0\"\n"
         # CMAKE_BUILD_TYPE
-        ret += "CMAKE_BUILD_TYPE=RelWithDebInfo\n"
+        if self.name == "catkin":
+            ret += "BUILD_BINARY=\"0\"\n"
```
tfoote: Is this equivalent to https://github.com/allenh1/superflore/pull/28/files#diff-5c266db61ea0f71dcf9a23aed337bb31L277

superflore/generators/ebuild/ebuild.py
```diff
@@ -250,9 +244,10 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
         ret += "\"\n\n"
 
         # SLOT
-        ret += "SLOT=\"{}\"\n".format(self.distro)
+        ret += "SLOT=\"0\"\n"
```
tfoote: We're moving toward a new mechanism to support parallel rosdistros?

superflore/generators/ebuild/ebuild.py
allenh1: @tfoote Yes, I set this up as an eclass variable which, if present, will set the `-DCATKIN_BUILD_BINARY_PACKAGE` flag to its value. Otherwise, that flag defaults to `1` (this just simplifies the catkin ebuild, making it more like the others).

superflore/generators/ebuild/ebuild.py
allenh1: Yes. In an email from @aballier, he recommended that I have different `USE` flags for each distro
(e.g. `ros_distro_lunar`), then have different slots for the versions of the packages. The next step will be to populate the `SLOT` field correctly.

PR #29, opened 2017-07-04
quoted description  string

Alessandro-Barbieri:
fixes https://github.com/allenh1/superflore/issues/3

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from shlex import quote
+try:
+    from shlex import quote
+except ImportError:
+    from pipes import quote
 from termcolor import colored
 import yaml
 import sys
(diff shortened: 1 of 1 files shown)


allenh1:
@Alessandro-Barbieri seems the CI doesn't like this.

allenh1:
Did you mean to pull in `quotes`?

```
shlex.quotes
   Characters that will be considered string quotes. The token accumulates until the same quote is 
   encountered again (thus, different quote types protect each other as in the shell.) By default, includes 
   ASCII single and double quotes.
```

If so, I don't think that's the fix here. I'd strongly prefer just a simple `.replace`.

Alessandro-Barbieri:
> Did you mean to pull in `quotes`?

No, python3's `shlex.quote` that was python2.7's  `pipe.quote`

```
shlex.quote(s)
 
Return a shell-escaped version of the string s. The returned value is a string that can safely be used as one token in a shell command line, for cases where you cannot use a list.
```
> I'd strongly prefer just a simple .replace.

Potentially this will let open some other vulnerabilities

https://xkcd.com/327/

^^

allenh1:
Ok, that's fine then.

Should these modules be there by default, or do they need to be added to the `setup.py` file?

Alessandro-Barbieri:
These modules are in the standard python library

PR #37, opened 2017-07-12
Add longdescription to metadata 84

### LINKED ISSUE #34: add longdescription to metadata
see https://devmanual.gentoo.org/ebuild-writing/misc-files/metadata/
TL;DR

```xml
<longdescription>
This is the longdescription
</longdescription>
```

allenh1:
Fix #34

### DIFF
--- a/superflore/generators/ebuild/metadata_xml.py
+++ b/superflore/generators/ebuild/metadata_xml.py
@@ -21,12 +21,16 @@ class metadata_xml(object):
         self.upstream_email = None
         self.upstream_bug_url = None
         self.maintainer_type = "person"
+        self.longdescription = "NONE"
 
     def get_metadata_text(self):
         ret = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
         ret += "<!DOCTYPE pkgmetadata SYSTEM "
         ret += "\"http://www.gentoo.org/dtd/metadata.dtd\">\n"
         ret += "<pkgmetadata>\n"
+        ret += "  <longdescription>\n"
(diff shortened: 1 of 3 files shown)


superflore/generators/ebuild/metadata_xml.py
```diff
@@ -21,12 +21,16 @@ def __init__(self):
         self.upstream_email = None
         self.upstream_bug_url = None
         self.maintainer_type = "person"
+        self.longdescription = "NONE"
```
tfoote: It would be cleaner to use None here instead of the string.

superflore/generators/ebuild/metadata_xml.py
```diff
@@ -21,12 +21,16 @@ def __init__(self):
...
 
     def get_metadata_text(self):
         ret = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
         ret += "<!DOCTYPE pkgmetadata SYSTEM "
         ret += "\"http://www.gentoo.org/dtd/metadata.dtd\">\n"
         ret += "<pkgmetadata>\n"
+        ret += "  <longdescription>\n"
```
tfoote: Then put these elements into in if block `if longdescription:` ...

superflore/generators/ebuild/run.py
```diff
@@ -31,6 +31,7 @@
 def link_existing_files():
     global overlay
     global mode
+    global preserve_existing
```
tfoote: I don't see this instance being used. And these globals are ugly. They should get cleaned up when swapping out argparse.

PR #38, opened 2017-07-13
quote instead of delete

Alessandro-Barbieri:
Deleting some characters lead to shitty looking descriptions.
Illegal characters can be reduced further to
```
"'`
```
only

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -126,7 +126,7 @@ class Ebuild(object):
         self.name = None
         self.has_patches = False
         self.die_msg = None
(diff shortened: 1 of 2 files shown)


allenh1:
Thanks for removing the `+-`... characters. Those definitely didn't look very good.

@tfoote Would you prefer the illegal characters be escaped, or would you prefer the illegal characters be removed?

My opinion here is that, since we only get 80 characters, I'd rather have more of the actual content than spend characters escaping the one's that aren't supposed to be in that field (note: they will still appear in the `metadata.xml` file's longdescription).

But if I'm the only one with that opinion, I'm open to just escaping them.

tfoote:
I think escaping will be easier to read, especially as the number of characters goes up. Packing in characters by dropping whitespace will only gain a marginal benefit of more characters and lose readability.

Issue #33, opened 2017-07-06
Implement argparse


Alessandro-Barbieri:
An option to force the regeneration of *all* packages will be nice.

allenh1:
That's what the `--all` flag does.

tfoote:
It looks like this option is available but hard to discover. @allenh1  I'd recommend that you use argparse instead of manually parsing the arguments. It's quite helpful and will provide more expected command line interactions.

allenh1:
That's probably a good idea. I'm too used to C, haha.

allenh1:
Fixed, #39

PR #39, opened 2017-07-16
Implemented argument parsing using argparse.

allenh1:
Connects to #33

### DIFF
--- a/superflore/generators/ebuild/metadata_xml.py
+++ b/superflore/generators/ebuild/metadata_xml.py
@@ -28,7 +28,8 @@ class metadata_xml(object):
         ret += "<!DOCTYPE pkgmetadata SYSTEM "
         ret += "\"http://www.gentoo.org/dtd/metadata.dtd\">\n"
         ret += "<pkgmetadata>\n"
-        if self.longdescription is not None:
+        if self.longdescription is not None and\
+           isinstance(self.longdescription, str):
             ret += "  <longdescription>\n"
             ret += "    " + self.longdescription + "\n"
             ret += "  </longdescription>\n"
--- a/superflore/generators/ebuild/metadata_xml.py
+++ b/superflore/generators/ebuild/metadata_xml.py
@@ -28,8 +28,7 @@ class metadata_xml(object):
         ret += "<!DOCTYPE pkgmetadata SYSTEM "
         ret += "\"http://www.gentoo.org/dtd/metadata.dtd\">\n"
         ret += "<pkgmetadata>\n"
-        if self.longdescription is not None and\
-           isinstance(self.longdescription, str):
+        if self.longdescription and isinstance(self.longdescription, str):
             ret += "  <longdescription>\n"
             ret += "    " + self.longdescription + "\n"
             ret += "  </longdescription>\n"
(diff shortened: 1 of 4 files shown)


superflore/generators/ebuild/metadata_xml.py
```diff
@@ -28,7 +28,8 @@ def get_metadata_text(self):
         ret += "<!DOCTYPE pkgmetadata SYSTEM "
         ret += "\"http://www.gentoo.org/dtd/metadata.dtd\">\n"
         ret += "<pkgmetadata>\n"
-        if self.longdescription is not None:
+        if self.longdescription is not None and\
```
tfoote: In python `is not None` is redundant with just `if args.distro:` it will also catch other zero values such as an empty string or list. https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none

Usually you only want to use the very specific meaning of None if you want to differentiate between other potential zero values.

superflore/generators/ebuild/run.py
```diff
@@ -69,33 +71,31 @@ def main():
...
-                ros_overlay.info(msg)
-                mode = arg1
-    elif len(sys.argv) >= 2:
-        ros_overlay.error('Invalid arguments!')
-        print_usage()
+
+    parser = argparse.ArgumentParser(
+        'Deploy ROS packages into Gentoo Linux.')
+    parser.add_argument(
+        '--distro', help='regenerate packages for the specified distro.', type=str)
```
tfoote: It would be better to use a default value here instead of putting the logic below. It's better for visibility as it will appear in the help. https://github.com/allenh1/superflore/pull/39/files#diff-5a7657ae13022a44815efc96706d2233R84

superflore/generators/ebuild/run.py
```diff
@@ -69,33 +71,31 @@ def main():
...
+    parser.add_argument(
+        '--all', help='regenerate all packages in all distros.')
+    args = parser.parse_args(sys.argv[1:])
+    download_yamls()
+    print(vars(args))
+    if args.distro is not None:
+        mode = args.distro
+    elif args.all is not None:
+        mode = 'all'
```
tfoote: Should this be mutating the same variable as the distro argument?

superflore/generators/ebuild/run.py
tfoote: This should probably be rosdistro for clarity.

superflore/generators/ebuild/metadata_xml.py
allenh1: That makes sense -- I think I have other places to change as well now...

allenh1: Thanks for the tip!

superflore/generators/ebuild/run.py
```diff
@@ -69,33 +71,31 @@ def main():
...
-        ros_overlay.error('Invalid arguments!')
-        print_usage()
+
+    parser = argparse.ArgumentParser(
+        'Deploy ROS packages into Gentoo Linux.')
+    parser.add_argument(
+        '--distro', help='regenerate packages for the specified distro.', type=str)
+    parser.add_argument(
+        '--all', help='regenerate all packages in all distros.')
+    args = parser.parse_args(sys.argv[1:])
```
allenh1: Also, pretty sure I shouldn't need to do the `sys.argv[1:]`, but argparse complained about invalid args otherwise. Any idea why?

superflore/generators/ebuild/run.py
allenh1: Yes -- I was being lazy. I'll fix that.

allenh1:
@tfoote No problem -- you are definitely allowed to be terse (and I didn't exactly help you out much with the timing of the PR).

It's certainly a lot cleaner now. Let me know if there's any more to do!

Issue #41, opened 2017-07-27
Docker Support


allenh1:
To allow generation on non-Gentoo platforms, it would be really nice to create a docker instance of
Gentoo, install repoman, and use a supplied pub key to generate the manifests.

https://hub.docker.com/u/gentoo/

@tfoote was able to spawn a quick Gentoo instance by using the following.

```
docker run -ti --rm gentoo/stage3-amd64
```

PR #44, opened 2017-07-28
Starting to implement docker integration.

allenh1:
connects to #41

### DIFF
--- a/repoman_docker/Dockerfile
+++ b/repoman_docker/Dockerfile
@@ -1,9 +1,10 @@
 FROM gentoo/stage3-amd64
 
 
-RUN emerge-webrsync && emerge --sync && emerge app-crypt/gnupg repoman
+RUN emerge-webrsync && emerge --sync && emerge app-crypt/gnupg repoman dev-vcs/git
 
 ENV FEATURES sign
 ENV PORTAGE_GPG_DIR ~/.gnupg/
 
-RUN echo '\nPORTDIR_OVERLAY="/usr/local/portage"\nPORTDIR=/usr/portage' >> /etc/portage/make.conf
+RUN mkdir -p /usr/local/portage
+RUN printf '\nPORTDIR_OVERLAY="/usr/local/portage"' >> /etc/portage/make.conf
--- a/repoman_docker/Dockerfile
+++ b/repoman_docker/Dockerfile
@@ -1,0 +1,9 @@
+FROM gentoo/stage3-amd64
+
+
+RUN emerge-webrsync && emerge --sync && emerge app-crypt/gnupg repoman
+
+ENV FEATURES sign
+ENV PORTAGE_GPG_DIR ~/.gnupg/
+
+RUN echo '\nPORTDIR_OVERLAY="/usr/local/portage"\nPORTDIR=/usr/portage' >> /etc/portage/make.conf
--- a/repoman_docker/Dockerfile
+++ b/repoman_docker/Dockerfile
@@ -1,7 +1,7 @@
 FROM gentoo/stage3-amd64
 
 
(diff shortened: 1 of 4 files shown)


repoman_docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM gentoo/stage3-amd64
+
+
+RUN emerge-webrsync && emerge --sync && emerge app-crypt/gnupg repoman dev-vcs/git
+
+ENV FEATURES sign
+ENV PORTAGE_GPG_DIR ~/.gnupg/
+
+RUN mkdir -p /usr/local/portage
+RUN printf '\nPORTDIR_OVERLAY="/usr/local/portage"' >> /etc/portage/make.conf
```
Alessandro-Barbieri: why not use `repos.conf`?

allenh1: I am just setting this up so that repoman is launched via a Docker container, so gentoo ebuilds can be generated on systems like Ubuntu.

Is there a point to configuring `repos.conf`?

Alessandro-Barbieri: https://wiki.gentoo.org/wiki//etc/portage/repos.conf#Custom_ebuild_repository

Alessandro-Barbieri: remember to run as root the commands or use another path

allenh1: > remember to run as root the commands or use another path

The docker container is already root when it runs, and all appears to be working atm.

superflore/generators/ebuild/overlay_instance.py
```diff
@@ -55,26 +56,21 @@ def commit_changes(self, distro):
         self.git.commit(m='{0}'.format(commit_msg))
 
     def regenerate_manifests(self, mode):
+        self.info('Createing docker instance...')
```
Alessandro-Barbieri: typo ^^

repoman_docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM gentoo/stage3-amd64
+
+
+RUN emerge-webrsync && emerge --sync && emerge app-crypt/gnupg repoman dev-vcs/git
```
Alessandro-Barbieri: consider adding `--autounmask-write` because portage is too interactive

superflore/generators/ebuild/overlay_instance.py
allenh1: Definitely. Thanks!

repoman_docker/Dockerfile
allenh1: Will do.

superflore/docker.py
```diff
@@ -0,0 +1,79 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from termcolor import colored
+from os import system
+import subprocess
+
+
+class docker(object):
```
tfoote: The class should be capitalized `Docker`

superflore/generators/ebuild/overlay_instance.py
```diff
@@ -55,26 +56,23 @@ def commit_changes(self, distro):
         self.git.commit(m='{0}'.format(commit_msg))
 
     def regenerate_manifests(self, mode):
+        self.info('Building docker image...')
+        dock = docker('repoman_docker', 'repoman')
```
tfoote: It would be better to use a more descriptive name for this image.

superflore/docker.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, docker_file, name):
```
tfoote: This isn't really  the "docker_file" but more of a dockerfile_directory.

Issue #50, opened 2017-08-01
traceback when running a 2nd time in a row


tfoote:
Related to #49 lets talk about how repeated builds and temp files might work better.

```
$ superflore-gen-ebuilds --dry-run --ros-distro lunar
Cloning repo ros/ros-overlay into directory /tmp/YmpJJukOup...
Creating new branch gentoo-bot-djqaSqVPxv...

Symbolicly linking files from /tmp/YmpJJukOup/ros-lunar...
Traceback (most recent call last):
  File "/home/tfoote/work/github/ros-infrastructure/superflore/superflore/generators/ebuild/run.py", line 92, in main
    link_existing_files(args.ros_distro)
  File "/home/tfoote/work/github/ros-infrastructure/superflore/superflore/generators/ebuild/run.py", line 43, in link_existing_files
    os.symlink(dir_fmt.format(overlay.repo_dir, mode), './ros-' + mode)
FileExistsError: [Errno 17] File exists: '/tmp/YmpJJukOup/ros-lunar' -> './ros-lunar'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/tfoote/work/github/ros-infrastructure/superflore/sfvenv/bin/superflore-gen-ebuilds", line 11, in <module>
    load_entry_point('superflore==0.0.0', 'console_scripts', 'superflore-gen-ebuilds')()
  File "/home/tfoote/work/github/ros-infrastructure/superflore/superflore/generators/ebuild/run.py", line 93, in main
    except os.FileExistsError:
AttributeError: module 'os' has no attribute 'FileExistsError'
(sfvenv) tfoote@snowman:~/work/github/ros-infrastructure/superflore Last: [1] (36s Seconds)
$ ll
total 88
drwxrwxr-x 11 tfoote tfoote  4096 Aug  1 16:30 ./
drwxrwxr-x 22 tfoote tfoote  4096 Aug  1 16:05 ../
-rw-rw-r--  1 tfoote tfoote  9849 Aug  1 16:27 abb_driver-None.tar.gz
-rw-rw-r--  1 tfoote tfoote  1388 Aug  1 16:27 abb-None.tar.gz
drwxrwxr-x  4 tfoote tfoote  4096 Aug  1 16:30 build/
drwxrwxr-x  2 tfoote tfoote  4096 Aug  1 16:30 dist/
drwxrwxr-x  8 tfoote tfoote  4096 Aug  1 16:05 .git/
[...]

Issue #32, opened 2017-07-06
option to choose existing overlay/folder


Alessandro-Barbieri:
I noticed that the repository get cloned in `/tmp` at every run but is sloooow. An option to specify an existing overlay will be great.

allenh1:
Could you propose a format for this option?

PR #51, opened 2017-08-01
Remove os prefix from consecutive build error.

allenh1:
Fixes traceback for #50 

Current behavior just removes the symlinks. I should adapt this to remove the existing tmp directory as well.

connects to #50 
connects to #32

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -14,6 +14,7 @@
 
 from superflore.generators.bitbake.gen_packages import generate_installers
 from superflore.generators.bitbake.ros_meta import ros_meta
+from superflore import RepoInstance
 import argparse
 import shutil
 import sys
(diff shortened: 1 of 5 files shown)


superflore/generators/bitbake/run.py
```diff
@@ -81,9 +81,8 @@ def main():
 
     # clone current repo
     selected_targets = active_distros
-    # try:
-    link_existing_files(args.ros_distro)
-    """
+    try:
+        link_existing_files(args.ros_distro)
```
tfoote: This looks to be linking args.ros_distro, but the exception removes all active distros?

tfoote: Related do we really need to symlink? Why don't we simple change the working directory/pass the repository location as an argument?

allenh1: > This looks to be linking args.ros_distro, but the exception removes all active distros?

The function attempts to remove all distros, and has a try catch around it that just passes if it fails to remove the link.

> Related do we really need to symlink? Why don't we simple change the working directory/pass the repository location as an argument?

I think that's a much better solution. I'm more partial to that behavior. Could you propose a format for that flag?

tfoote: This goes with the ability to reuse an exisiting repo. I'd suggest `--output-repository-path` as an argument. Where if it's unset it will use a temporary directory. (If it uses a temporary directory it should delete it at the end after opening the PR) There could be a flag to prevent the cleanup. and logic to prevent you from running without a PR or preventing cleanup.

allenh1: Sounds good to me!

allenh1:
@tfoote I added the `--output-repository-path` option.

tfoote:
Adding the output repository path is good. But I'd like to go one step further and not create the symlinks and just work in the repositories directory.

allenh1:
@tfoote I've removed the sym-link logic, and just rebased on top of master.

Issue #49, opened 2017-08-01
Suggestions for improved usage


tfoote:
Please add to the readme usage documentation. I found some executables by tab completing. However there should be documentation of the top level commands.

Here's some more features for better usability.

I ran with --dry-run and it finished w/o filing a PR, but didn't tell me where the output was. (I found it in /tmp by listing latest changes.) If it generates a temp-dir it should clean it up or at least tell you to clean up.

It would be great to parameterize the output directory to allow reuse of a preexisting local checkout of the repo so you don't have to refetch it if you already have a clone.

Following that if you run with --dry-run and a non-temp directory an option --pr-only would be helpful after reviewing the result.

And adding arguments for the upstream repo will be important for anyone who wants to fork etc. Also supporting a config file would be good for this too.

allenh1:
> I ran with --dry-run and it finished w/o filing a PR, but didn't tell me where the output was. (I found it in /tmp by listing latest changes.) If it generates a temp-dir it should clean it up or at least tell you to clean up.

That's very true. I'll add a message to the end of the dry run that gives the temp directory and says that they should clean it up or run again with `--pr-only`.

> It would be great to parameterize the output directory to allow reuse of a preexisting local checkout of the repo so you don't have to refetch it if you already have a clone.

I think @Alessandro-Barbieri suggested something similar in #32. I'll add that to the list.

> Also supporting a config file would be good for this too.

Sounds good to me. Things like `overlay-repo`, `overlay-branch`, `ros-distro`, are some possible settings I'm thinking of. Got any more for me?

allenh1:
@tfoote I think this is resolved -- can this be closed?

tfoote:
It's fixed for ebuilds. Do you want to track adding open embedded separately?

tfoote:
Actually reading it through again. A high level of what the process does and what expected inputs and outputs are would be very helpful.

allenh1:
> And adding arguments for the upstream repo will be important for anyone who wants to fork etc. Also supporting a config file would be good for this too.

@tfoote I think this is the only residual complaint for this issue, right? Just going through the open issues. I also have some of these changes to replicate for open embedded.

tfoote:
yeah, that new argument should be kept in mind but things are much better now.

PR #52, opened 2017-08-02
Begin implementing the pr-only option.

allenh1:
connects to #49

### DIFF
--- a/superflore/generators/bitbake/ros_meta.py
+++ b/superflore/generators/bitbake/ros_meta.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 
-from superflore import repo_instance
+# TODO(allenh1): remove inheritance like the RosOverlay class.
+from superflore import RepoInstance as repo_instance
 import random
 import string
 import time
--- a/superflore/generators/ebuild/overlay_instance.py
+++ b/superflore/generators/ebuild/overlay_instance.py
@@ -1,5 +1,5 @@
 # Instance of the ROS Overlay
-from superflore import repo_instance
+from superflore import RepoInstance
 from superflore.docker import Docker
 import random
 import string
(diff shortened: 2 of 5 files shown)


allenh1:
I think I'll catch the OpenEmbedded args up at a later date.

superflore/generators/ebuild/overlay_instance.py
```diff
@@ -19,14 +19,14 @@ def get_random_branch_name():
 
 
 class ros_overlay(repo_instance):
-    def __init__(self):
+    def __init__(self, existing_repo=None):
```
tfoote: Is this new argument used?

superflore/repo_instance.py
```diff
@@ -19,7 +19,7 @@
 
 
 class repo_instance(object):
```
tfoote: This class should be RepoInstance

superflore/generators/ebuild/overlay_instance.py
```diff
@@ -19,14 +19,14 @@ def get_random_branch_name():
 
 
 class ros_overlay(repo_instance):
```
tfoote: This class should be CamelCased as well.

superflore/generators/ebuild/overlay_instance.py
```diff
@@ -19,14 +19,14 @@ def get_random_branch_name():
...
-    def __init__(self):
+    def __init__(self, existing_repo=None):
         # clone repo into a random tmp directory.
         repo_instance.__init__(self, 'ros',
```
tfoote: Use super to init the base class. lots of details here: https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods

superflore/generators/ebuild/overlay_instance.py
tfoote: Looking at the initialization. Do you really need the inheritance here or can you just have the repo_instance as a member?

allenh1: That's a good point.

I think I should also move the `pull_request` function into `RepoInstance`

superflore/generators/ebuild/overlay_instance.py
allenh1: It would appear not -- good catch.

superflore/generators/bitbake/ros_meta.py
```diff
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 
-from superflore import repo_instance
+# TODO(allenh1): remove inheritance like the RosOverlay class.
+from superflore import RepoInstance as repo_instance
```
allenh1: This is a temporary hacky-fix to avoid breaking things in the OpenEmbedded world. This will be updated when I catch OpenEmbedded up.

allenh1:
@tfoote rebased on top of master. Ready for review.

PR #53, opened 2017-08-03
Adds a way to remove Python3 support from packages

allenh1:
It's not great, but I can't think of a truly robust way to do this.

### DIFF
--- a/superflore/generators/ebuild/gen_packages.py
+++ b/superflore/generators/ebuild/gen_packages.py
@@ -31,7 +31,7 @@ org = "Open Source Robotics Foundation"
 org_license = "BSD"
 
 """
-This is a blacklist of things that
+TODO(allenh1): This is a blacklist of things that
 do not yet support Python 3. This will
 be updated on an as-needed basis until
 a better solution is found (CI?).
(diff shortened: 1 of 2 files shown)


superflore/generators/ebuild/gen_packages.py
```diff
@@ -30,6 +30,13 @@
 org = "Open Source Robotics Foundation"
 org_license = "BSD"
 
+"""
+This is a blacklist of things that
```
tfoote: Please turn this into a comment with a TODO in it to make it easier to find in the future. This should be something that goes into a config file somewhere in the future.

superflore/generators/ebuild/gen_packages.py
```diff
@@ -30,6 +30,13 @@
 org = "Open Source Robotics Foundation"
 org_license = "BSD"
 
+"""
+TODO(allenh1): This is a blacklist of things that
```
tfoote: Triple quotes aren't comments they're multi line strings that are evaluated by the interpreter.

allenh1: > formatting the comment as a comment

I see what you meant now...

Issue #35, opened 2017-07-07
trim DESCRIPTION to 80 chars


Alessandro-Barbieri:
as title say
bonus: trim on whitespace
bonus: add ` [...]` at the end (respecting the 80 characters limit)

allenh1:
@Alessandro-Barbieri is this something you're working on?

Alessandro-Barbieri:
Not now, feel free to do what you want

PR #60, opened 2017-08-10
Trim description to 80, and end with '[...]'

allenh1:
This cleans up the truncation a little better.

connects to #35

### DIFF
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -63,6 +63,17 @@ def sanitize_string(string, illegal_chars):
     return ret
 
 
+def trim_string(string, length=80):
+    if len(string) < length:
+        return string
+    end_string = '[...]'
+
+    return '{0}{1}'.format(
+        string[:length - len(end_string)],
+        end_string
+    )
+
+
 def get_license(l):
     bsd_re = '^(BSD)((.)*([1234]))?'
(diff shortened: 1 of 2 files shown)


superflore/utils.py
```diff
@@ -63,6 +63,17 @@ def sanitize_string(string, illegal_chars):
     return ret
 
 
+def trim_string(string, length=80):
+    if len(string) < length:
+        return string
+    end_string = '[...]'
+
+    return '{0}{1}'.format(
+        string[:length - len(end_string)],
```
tfoote: You don't need to bother with the format here. You can just add the strings. 

`return string[:length - len(end_string)] + end_string`

superflore/utils.py
```diff
@@ -63,6 +63,15 @@ def sanitize_string(string, illegal_chars):
     return ret
 
 
+def trim_string(string, length=80):
+    if len(string) < length:
+        return string
+    end_string = '[...]'
+
+    return string[:length - len(end_string)]\
```
tfoote: Don't wrap here, it's only 57 characters total. One line is much more readable.

allenh1: Got it -- fixed.

Alessandro-Barbieri:
thanks

allenh1:
Sure thing!

PR #69, opened 2017-08-28
Remove Extra Newlines

allenh1:
This should work -- I'll need to test it, since there were no updates today (strange!).


allenh1:
@tfoote tested in ros/ros-overlay#166 and looks to be working.

PR #73, opened 2017-10-19
Add test_utils.py file to start implementing tests.

allenh1:
Tests:
---------
 * [x] `sanitize_string`
 * [x] `trim_string`
 * [x] `get_license`

Documentation:
----------------------
 * [ ] `sanitize_string`
 * [ ] `trim_string`
 * [ ] `get_license`
 * [ ] `resolve_dep`

### DIFF
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -71,44 +71,54 @@ def trim_string(string, length=80):
 
 
 def get_license(l):
-    bsd_re = '^(BSD)((.)*([1234]))?'
-    gpl_re = '^(GPL)((.)*([123]))?'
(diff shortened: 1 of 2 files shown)


superflore/utils.py
```diff
@@ -70,44 +70,54 @@ def trim_string(string, length=80):
 
 
 def get_license(l):
-    bsd_re = '^(BSD)((.)*([1234]))?'
-    gpl_re = '^(GPL)((.)*([123]))?'
```
tfoote: Without this will it overmatch on a LGPL?

allenh1: Definitely would have... I just went ahead and swapped the order.

PR #74, opened 2017-10-19
Fix Nosetests invocation in CI

allenh1:
More python3/python2 confusion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,3 @@ install:
 # command to run tests
 script:
   - python -m 'nose'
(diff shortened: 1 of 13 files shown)


allenh1:
Currently, I have no idea how to force travis to run flake8 for python3 and not ruin the test file for python2.

This currently just fixes the linter errors I get when I run it locally.

allenh1:
I think we'll just stick with python2 flake8 and handle the python3 flake8 problems (which seem to only reflect import orders) on a when-noticed basis.

allenh1:
Nevermind, turns out I was missing a [plugin for flake8.](https://pypi.python.org/pypi/flake8-import-order)

PR #76, opened 2017-10-23
Add `--only PKG` option

allenh1:
This set of changes adds an option to generate a single package by name, and will also include a refactor of `generate_packages.py` so that it can be relocated outside the `ebuild` folder (this will eliminate a lot of redundancy).

ToDo:
---------
 * [x] add the `--only PKG` flag to generate a single package
 * [x] refactor & separate gentoo-specific logic
 * [x] modify `regenerate_manifests` to only regenerate changed packages

### DIFF
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -64,7 +64,7 @@ def generate_installers(
             ok('{0}%: {1} \'{2}\'.'.format(percent, success_msg, pkg))
             succeeded = succeeded + 1
             changes.append('*{0} --> {1}*'.format(pkg, version))
-            installers.append(current)
+            installers.append(pkg)
         except KeyError:
             failed_msg = 'Failed to generate installer'
             err("{0}%: {1} for package {2}!".format(percent, failed_msg, pkg))
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -1,0 +1,79 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from rosinstall_generator.distro import get_distro
+from rosinstall_generator.distro import get_package_names
+
+from superflore.exceptions import UnresolvedDependency
+
+from superflore.utils import err
+from superflore.utils import get_pkg_version
+from superflore.utils import ok
+from superflore.utils import warn
+
+
+def generate_installers(distro_name, overlay, gen_pkg, preserve_existing=True):
+    distro = get_distro(distro_name)
+    pkg_names = get_package_names(distro)
+    total = float(len(pkg_names[0]))
+    borkd_pkgs = dict()
+    changes = []
+    installers = []
+    bad_installers = []
+    succeeded = 0
+    failed = 0
+
+    for i, pkg in enumerate(sorted(pkg_names[0])):
+        version = get_pkg_version(distro, pkg)
+        ebuild_name =\
+            '/ros-{0}/{1}/{1}-{2}.ebuild'.format(distro_name, pkg, version)
+        ebuild_name = overlay.repo.repo_dir + ebuild_name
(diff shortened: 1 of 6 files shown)


allenh1:
@tfoote I'm working on this right now before I look into more unit tests and making the yocto stuff better in an effort to reduce the duplicate sections of code.

My line of thought here is to make a `gen_package_list` function that the entry points call. This function should return the a list of the packages to be built, together with metadata. This list gets passed to an entry point specific target, and the target does with it whatever it's supposed to do (so generating ebuilds and generating manifests, or bitbake stuff).

I think this would also make things better in the future for extending things, and I can't figure out what went wrong in my head wherein I didn't write it this way the first time.

Does this sound like a good step to you?

allenh1:
Rebased on top of #77

allenh1:
@tfoote I was going to remove the Open Embedded dependency with this PR, then I noticed how many commits in I was. I'll adapt there after this one, I think.

allenh1:
@wjwwood

allenh1:
@tfoote

allenh1:
THIS IS MY TIME TO SHINE. YOU ARE NOT ALLOWED BY LAW TO DISCRIMINATE.

allenh1:
I'm so sorry for anyone looking at this -- my account has been hacked.

superflore/generate_installers.py
```diff
@@ -0,0 +1,84 @@
...
+            installers.append(pkg)
+        except KeyError:
+            failed_msg = 'Failed to generate installer'
+            err("{0}%: {1} for package {2}!".format(percent, failed_msg, pkg))
+            bad_installers.append(pkg)
+            failed = failed + 1
+    results = 'Generated {0} / {1}'.format(succeeded, failed + succeeded)
+    results += ' for distro {0}'.format(distro_name)
+    info("------ {0} ------".format(results))
+    print()
```
tfoote: It's weird to use print in addition to the logging functions. Maybe just add a newline to the above?

superflore/generate_installers.py
```diff
@@ -0,0 +1,84 @@
...
+from superflore.utils import info
+from superflore.utils import ok
+from superflore.utils import warn
+
+
+def generate_installers(
+    distro_name,      # ros distro name
+    overlay,          # repo instance
+    gen_pkg_func,     # function to call for generating
+    update=True       # are we in update mode?
```
tfoote: What does update mean here? I thought it was forcing an update but https://github.com/ros-infrastructure/superflore/pull/76/files#diff-22c1a4ad79cb71ae0b51b9ef9621bf80R60 doesn't make sense in that case.

superflore/generators/ebuild/gen_packages.py
```diff
@@ -42,145 +44,86 @@
...
-
-
-def err(string):
-    print(colored('!!!! {0}'.format(string), 'red'))
+org = "Open Source Robotics Foundation"
+org_license = "BSD"
 
 
-def make_dir(dirname):
+def regenerate_pkg(overlay, pkg, distro_name=None, distro=None, update=False):
```
tfoote: Do you really need both distro and distro_name support? That seems like it is just going to be more to test. Since it's only a one line command to get the distro. It would be better to do that from the calling instance and not have the ambiguous API.

superflore/generators/ebuild/run.py
```diff
@@ -82,29 +90,34 @@ def main():
         help='location of the Git repo',
         type=str
     )
+    parser.add_argument(
+        '--only',
+        help='generate only the specified package',
+        type=str
```
tfoote: I'd recommend supporting mutiple listed packages. You can use `nargs='+'` https://docs.python.org/3/library/argparse.html#nargs

Note you'll slightly need to change the logic https://github.com/ros-infrastructure/superflore/pull/76/files#diff-5a7657ae13022a44815efc96706d2233R147 to iterate over the multiple arguments.

superflore/generate_installers.py
allenh1: Ah, `update` gets set to `True` to keep the ebuilds that are currently there (so quite the opposite). I'll change it back to `preserve_existing`, which it was previously.

superflore/generate_installers.py
allenh1: Ya, that's definitely true. I might just remove it all together.

superflore/generators/ebuild/gen_packages.py
allenh1: > Since it's only a one line command to get the distro. It would be better to do that from the calling instance and not have the ambiguous API.

Ya, that's true. I was going for an optimization here, but I don't think I really saved myself anything.

superflore/generators/ebuild/run.py
allenh1: I was wondering how one would go about this. ;) Thanks for the resources. I'll definitely add this in.

allenh1:
@tfoote Ok, ready for round two.

PR #77, opened 2017-10-23
Fix Linter Errors

allenh1:
New linter errors are showing up on master (I guess Flake8 updated?).

### DIFF
--- a/superflore/generators/bitbake/gen_packages.py
+++ b/superflore/generators/bitbake/gen_packages.py
@@ -56,7 +56,7 @@ def info(string):
 def make_dir(dirname):
     try:
         os.makedirs(dirname)
-    except:
+    except Exception:
         pass
 
 
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -93,7 +93,7 @@ def main():
             warn_msg =\
                 'removing existing symlink "./recipes-ros-{0}"'.format(x)
             RepoInstance.warn(warn_msg)
-        except:
+        except Exception:
             pass
     link_existing_files(args.ros_distro)
 
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -93,8 +93,11 @@ def main():
             warn_msg =\
                 'removing existing symlink "./recipes-ros-{0}"'.format(x)
             RepoInstance.warn(warn_msg)
-        except Exception:
-            pass
+        except OSError as e:
+            if e.errno == errno.ENOENT:
+                pass
+            else:
+                raise e
     link_existing_files(args.ros_distro)
 
     # generate installers
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import argparse
+import errno
 import os
 import shutil
 import sys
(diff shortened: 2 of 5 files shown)


superflore/generators/ebuild/ebuild.py
```diff
@@ -142,9 +142,9 @@ def get_ebuild_text(self, distributor, license_text, die_msg=None):
                     ret += get_license(self.upstream_license) + "\"\n\n"
             elif isinstance(self.upstream_license, list):
                 ret += "LICENSE=\"( "
-                for l in self.upstream_license:
-                    l = get_license(l)
-                    ret += '{0} '.format(l)
+                for _ul in self.upstream_license:
```
tfoote: This doesn't need a for loop, use list comprehension.

```
ret += ' '.join([get_license(ul) for ul in self.upstream_license])
```

allenh1: That's a good tip -- thanks!

superflore/generators/bitbake/gen_packages.py
```diff
@@ -56,7 +56,7 @@ def info(string):
 def make_dir(dirname):
     try:
         os.makedirs(dirname)
-    except:
+    except Exception:
```
tfoote: I'd recommend actually checking the errno since already present is valid but if you're out of disk space etc it should fail: https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python

superflore/generators/ebuild/run.py
```diff
@@ -102,15 +102,15 @@ def main():
                 msg = msg_file.read().rstrip('\n')
             with open('.pr-title.tmp', 'r') as title_file:
                 title = title_file.read().rstrip('\n')
-        except:
+        except Exception as e:
```
tfoote: Is there any useful info inside the exception e that could be shown?

superflore/generators/ebuild/gen_packages.py
```diff
@@ -58,7 +58,7 @@ def err(string):
 def make_dir(dirname):
     try:
         os.makedirs(dirname)
-    except:
+    except Exception:
```
tfoote: This is repeated as above. Is there somewhere it can be shared?

superflore/generators/bitbake/run.py
```diff
@@ -93,7 +93,7 @@ def main():
             warn_msg =\
                 'removing existing symlink "./recipes-ros-{0}"'.format(x)
             RepoInstance.warn(warn_msg)
-        except:
+        except Exception:
```
tfoote: What sort of exceptions are ok here vs not ok? This satisfies the linter but still doesn't resolve the underlying question

superflore/generators/bitbake/gen_packages.py
allenh1: Since I'm centralizing this function in #76, I'm just going to fix that there (26f99eb1571800f85da2ba428289e8e59cc12531).

superflore/generators/ebuild/gen_packages.py
allenh1: Response as above ;)

allenh1:
@tfoote Ok, ready for round three!

superflore/generators/bitbake/gen_packages.py
tfoote: +1

Issue #65, opened 2017-08-11
Add stdeb.cfg to support building debian packages


tfoote:
Here's an example: https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg

We talked about only targeting python3

Make sure to call out git explicitly as a dependency. I ran into that here: https://gist.github.com/tfoote/4fff504f608ad45cf3e897cd2186025b

allenh1:
Resolved by #133.

PR #78, opened 2017-10-31
Add stdeb.cfg

allenh1:
@tfoote There's a couple of dependencies I'm not sure how to handle.

* `python3-docker`: the pip version is 2.5, and the Ubuntu version is 1.9. You mentioned they were pretty different? Should I be concerned?
* `git-pull-request`: the python library lacks a `.deb` installer. Do I mark it as a pip installed thing somehow?

connects to ros-infrastructure/superflore#65


tfoote:
> python3-docker: the pip version is 2.5, and the Ubuntu version is 1.9. You mentioned they were pretty different? Should I be concerned?

Potentially. That is a pretty big delta. 

> git-pull-request: the python library lacks a .deb installer. Do I mark it as a pip installed thing somehow?

That's going to be a problem. Dependencies of debian packages must be debian packages.

allenh1:
> That's going to be a problem. Dependencies of debian packages must be debian packages.

That's what I was afraid of... Is the next step then to manufacture packages for those dependencies, or just say that people should pip install them?

tfoote:
Lets let this sit in the background and use source or pip installations for now.

allenh1:
Closing for now.

allenh1:
@tfoote Since we've removed the `git-pull-request` module, I think we can consider this again.

Issue #62, opened 2017-08-11
Update bitbake generator to not use symlinks


tfoote:
As implemented in ebuild generator in #51

allenh1:
Fixed in #79.

PR #79, opened 2017-10-31
Feature Parity for Open Embedded

allenh1:
This is the Open Embedded side of #76's changes.

### DIFF
--- a/superflore/generators/bitbake/ros_meta.py
+++ b/superflore/generators/bitbake/ros_meta.py
@@ -18,7 +18,12 @@ import random
 import string
 import time
 
-from superflore import RepoInstance as repo_instance
+from superflore.repo_instance import RepoInstance as repo_instance
+
+from superflore.utils import err
+from superflore.utils import info
+from superflore.utils import ok
+from superflore.utils import warn
 
 
 def get_random_tmp_dir():
@@ -28,7 +33,7 @@ def get_random_tmp_dir():
 
 def get_random_branch_name():
     rand_str = ''.join(random.choice(string.ascii_letters) for x in range(10))
-    return 'gentoo-bot-{0}'.format(rand_str)
+    return 'yocto-bot-{0}'.format(rand_str)
 
 
 class ros_meta(repo_instance):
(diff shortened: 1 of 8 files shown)


allenh1:
Commit 2a24583 resolves #62.

superflore/generators/bitbake/ros_meta.py
```diff
@@ -12,67 +12,63 @@
...
 
-
-# TODO(allenh1): remove inheritance like the RosOverlay class.
+import os
 import random
 import string
 import time
 
-from superflore import RepoInstance as repo_instance
+from superflore.generators.ebuild.overlay_instance import get_random_tmp_dir
```
tfoote: This looks like a chance for better sharing by moving to utils?

superflore/generators/bitbake/ros_meta.py
```diff
@@ -12,67 +12,63 @@
...
+class RosMeta(object):
+    def __init__(self, repo_dir=None):
         # clone repo into a random tmp directory.
-        repo_instance.__init__(self, 'allenh1',
-                               'meta-ros', get_random_tmp_dir())
+        do_clone = True
+        if repo_dir:
+            do_clone = not os.path.exists(os.path.realpath(repo_dir))
+        self.repo = RepoInstance(
+            'allenh1', 'meta-ros', repo_dir or get_random_tmp_dir(),
```
tfoote: hard coded value here, at least make it an argument with a default

superflore/generators/bitbake/ros_meta.py
allenh1: Yes, this was a consideration... I think I'll change it to have a function that returns a group of random characters of some length. That way I could just call it and inline the formatted directories.

tfoote: Sorry looking under the hood you should be using tempfile.mkdtemp And while you're doing that use a context manager to clean it afterwards. like this: https://gist.github.com/tfoote/f68ad3f6a6a04a41264f65a04ef86fac

allenh1:
@tfoote I'm a little confused as to how I implement the TempfileManager. I understand I'm supposed to use a `with` statement, but I'm not sure where the `with` should go... Any insight here?

Evidlo:
@allenh1 See [this](https://stackoverflow.com/questions/1984325/explaining-pythons-enter-and-exit)

tfoote:
> @tfoote I'm a little confused as to how I implement the TempfileManager. I understand I'm supposed to use a with statement, but I'm not sure where the with should go... Any insight here?

Instead of doing 
```
mkdir foo
operate inside foo
rmdir foo
```
change it to:
```
with TempfileManager(foo) as tempdir:
    operate inside foo
```
When the with statement is finished the __exit__ method will be called and things will be cleaned up. Likely you'll want this in your top level logic (near the command line options) and the working directory will be passed into the build specific logic which won't know if it's temporary or passed via an argument to avoid cleanup. And they can use a implementation specific subdirectory of the tempdir.

allenh1:
@tfoote nice tip! Took me a few to understand it, but it's super neat.

Issue #80, opened 2017-11-02
Clean up .tar files for OpenEmbedded


allenh1:
Currently, the tar files for OpenEmbedded are just left to the user to clean up, which is far less than desirable.

This is a long-standing issue that was to be fixed in #79, but, due to the current length of that PR, has been deferred to later.

tfoote:
Note that if you do this inside of the temporary directory they will automatically get cleaned up.

PR #81, opened 2017-11-02
Remove tar file after use.

allenh1:
This was much easier than I thought it was going to be.

Connects to #80

### DIFF
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -23,10 +23,11 @@ from superflore.utils import warn
 
 
 def generate_installers(
-    distro_name,            # ros distro name
-    overlay,                # repo instance
-    gen_pkg_func,           # function to call for generating
-    preserve_existing=True  # don't regenerate if installer exists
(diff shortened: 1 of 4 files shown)


allenh1:
@tfoote I suppose I could have put this in the other PR now that I see the diff. Was expecting something to go wrong, but this seems to work.

allenh1:
Ok, that should be good now. Ready for review.

tfoote:
A way to preserve the archive for repeated runs would be important to have to save time, instead of requiring a full redownload for every run.

allenh1:
@tfoote Maybe a `--save-tars DIR` flag would be appropriate?

tfoote:
I'd suggest something like `--tar-archive-dir` where that directory specifies where to build the archive and it also won't be cleaned up. You can reuse the TempfileManager for this, it's the same logic.

allenh1:
@tfoote Ok, I think this is ready now.

PR #88, opened 2017-11-09
Add an install for the dockerfile.

allenh1:
This _should_ tell setup.py to install the dockerfile (so that we can have it in a central location and not run superflore from the source directory).

But I'm not sure where the file goes exactly?

> Each (directory, files) pair in the sequence specifies the installation directory and the files to install there. If directory is a relative path, it is interpreted relative to the installation prefix (Python’s sys.prefix for pure-Python packages, sys.exec_prefix for packages that contain extension modules). Each file name in files is interpreted relative to the setup.py script at the top of the package source distribution. No directory information from files is used to determine the final location of the installed file; only the name of the file is used.

This is from the [Python documentation](https://docs.python.org/3/distutils/setupscript.html#installing-additional-files).

@tfoote Do you know how one goes about referencing these files/if this is a proper way to do this?

### DIFF
--- a/superflore/docker.py
+++ b/superflore/docker.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 import os
-import docker
 
+import docker
 from superflore.utils import info
 from superflore.utils import ok
 
--- a/superflore/docker.py
+++ b/superflore/docker.py
@@ -20,9 +20,9 @@ from superflore.utils import ok
 
 
 class Docker(object):
-    def __init__(self, dockerfile_directory, name):
+    def __init__(self, dockerfile, name):
         self.client = docker.from_env()
-        self.dockerfile_directory = os.path.dirname(dockerfile_directory)
+        self.dockerfile_directory = os.path.dirname(dockerfile)
         self.name = name
         self.image = None
         self.directory_map = dict()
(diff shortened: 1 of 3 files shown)


tfoote:
You want to use the pkg_resources resource manager api: https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resourcemanager-api

allenh1:
@tfoote Ready for review!

superflore/docker.py
```diff
@@ -20,7 +22,7 @@
 class Docker(object):
     def __init__(self, dockerfile_directory, name):
         self.client = docker.from_env()
-        self.dockerfile_directory = dockerfile_directory
+        self.dockerfile_directory = os.path.dirname(dockerfile_directory)
```
tfoote: You're passing in the filename but calling it dockerfile_directory.  You should either change the variable name or pass in the dirname only and not run the dirname here.

allenh1: Good point. I think I'll change the parameter name to be `dockerfile`, and then open it's directory and set it to `self.dockerfile_directory`.

tfoote:
Note when you squash, I'd suggest editing down the commit message as many of the in progress commits comment are not relevant after merging.

PR #100, opened 2017-12-07
Replicated copyright change to open embedded, and resolved multi-license ToDo

shaneallcroft:
#100 Hooray!!!!!!

### DIFF
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -133,9 +133,7 @@ class yoctoRecipe(object):
         ret += 'SECTION = "devel"\n'
         self.get_license_line()
         if isinstance(self.license, str):
-            self.license = self.license.split(',')[0]
-            ret += 'LICENSE = "'
-            ret += ' & '.join([get_license(l) for l in self.license]) + '"\n'
+            ret += 'LICENSE = "%s"\n' % getlicense(self.license)
         elif isinstance(self.license, list):
             ret += 'LICENSE = "'
             ret += ' & '.join([get_license(l) for l in self.license]) + '"\n'
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -136,9 +136,7 @@ class yoctoRecipe(object):
             self.license = self.license.replace(' ', '-')
             ret += 'LICENSE = "' + self.license + '"\n'
         elif isinstance(self.license, list):
-            self.license = self.license[0].replace(' ', '-')
-            ret += 'LICENSE = "' + self.license + '"\n'
-            ret += ' '.join(self.license)
+            ret += 'LICENSE = "' + ' & '.join(self.license)
             ret += '"\n'
         ret += 'LIC_FILES_CHKSUM = "file://package.xml;beginline='
         ret += str(self.license_line)
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -136,7 +136,7 @@ class yoctoRecipe(object):
             self.license = self.license.replace(' ', '-')
             ret += 'LICENSE = "' + self.license + '"\n'
         elif isinstance(self.license, list):
-            ret += 'LICENSE = "' + ' & '.join(self.license)
+            ret += 'LICENSE = "' + ' & '.join([getlicense(l) for l in self.license])
             ret += '"\n'
         ret += 'LIC_FILES_CHKSUM = "file://package.xml;beginline='
         ret += str(self.license_line)
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -33,6 +33,7 @@ from superflore.exceptions import NoPkgXml
 from superflore.utils import get_pkg_version
 from superflore.utils import info
 from superflore.utils import resolve_dep
+from superflore.utils import getlicense
 
 
 class yoctoRecipe(object):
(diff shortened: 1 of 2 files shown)


shaneallcroft:
@tfoote @allenh1

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -136,16 +138,8 @@ def get_recipe_text(self, distributor, license_text, die_msg=None):
...
             ret += 'LICENSE = "' + self.license + '"\n'
-            """
-            TODO(allenh1): add this functionality
-            first = True
-            for lic in self.license:
-                if not first:
-                    ret += ' '
-                    first = False
-                ret += lic
+            ret += ' '.join(self.license)
```
tfoote: This prints the licences twice, once in a list format and once joined by strings. It's probably worth removing the above print and joining on ', ' for human readability. @allenh1 I assume this is optimized for human readability so comma separated list is best?

allenh1: @tfoote I'm actually not sure how Open Embedded wants multiple licenses to be handled... I think I'll just pull this change locally and try it out? I'm honestly unsure.

tests/ebuild/simple_expected.ebuild
```diff
@@ -8,7 +8,7 @@ inherit ros-cmake
 
 DESCRIPTION="an ebuild"
 HOMEPAGE="https://www.website.com"
-SRC_URI="https://www.website.com/download/${PN}/archive/${PN}/release/lunar/0.0.0.tar.gz -> ${PN}-lunar-release-${PV}.tar.gz"
+SRC_URI="https://www.website.com/download/stuff.tar.gz -> ${PN}-lunar-release-${PV}.tar.gz"
```
allenh1: This change shouldn't be made -- perhaps this got pulled in during the rebase somehow?

shaneallcroft: I'll revert that now :+1:

allenh1:
@shaneallcroft Ok, there's an odd way that OE wants this handled.

```
LICENSE = "GPLv2 & LGPLv2 & BSD & MIT"
```

Shouldn't be that difficult to adapt to, but lmk if you run into any issues.

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -136,16 +138,8 @@ def get_recipe_text(self, distributor, license_text, die_msg=None):
         elif isinstance(self.license, list):
             self.license = self.license[0].replace(' ', '-')
```
allenh1: @shaneallcroft this hackyness needs to be killed with fire. This just removes the list entirely and replaces it with its first element.

Therefore, this line and the one that immediately follows should be removed.

shaneallcroft: You got it! :+1:

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -134,18 +136,8 @@ def get_recipe_text(self, distributor, license_text, die_msg=None):
...
-            ret += 'LICENSE = "' + self.license + '"\n'
-            """
-            TODO(allenh1): add this functionality
-            first = True
-            for lic in self.license:
-                if not first:
-                    ret += ' '
-                    first = False
-                ret += lic
+            ret += 'LICENSE = "' + ' & '.join(self.license)
```
allenh1: This is very close. Make sure to call the `getlicense` function on each entry of the list! Something like the following would probably do the trick.

```py
' & '.join([getlicense(l) for l in self.license])
```

shaneallcroft: OH yeah, that makes sense, cause the hackyness is gone so every item needs to be taken into account im on it :+1:

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -134,18 +136,8 @@ def get_recipe_text(self, distributor, license_text, die_msg=None):
             self.license = self.license.replace(' ', '-')
             ret += 'LICENSE = "' + self.license + '"\n'
```
allenh1: @shaneallcroft You might update this one as well.

shaneallcroft: Oh SNAP i gotchu

shaneallcroft:
:( wait i messed it up

allenh1:
> :( wait i messed it up

It happens. Looks like you just missed an import for the `getlicense` function.

shaneallcroft:
I forgot the underscore i think just a sec

shaneallcroft:
First try :sunglasses:

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -131,21 +134,11 @@ def get_recipe_text(self, distributor, license_text, die_msg=None):
         self.get_license_line()
         if isinstance(self.license, str):
             self.license = self.license.split(',')[0]
-            self.license = self.license.replace(' ', '-')
-            ret += 'LICENSE = "' + self.license + '"\n'
+            ret += 'LICENSE = "'
+            ret += ' & '.join([get_license(l) for l in self.license]) + '"\n'
         elif isinstance(self.license, list):
```
allenh1: So I don't think you need to use a join when `self.license` is a string. I think this line could just be

```py
ret += 'LICENSE = "%s"\n' % getlicense(self.license)
```

shaneallcroft: Ahhh, i didnt see that '[0]' on line 136, thought it was just being made into a list there so i treated it the same, my b!~

shaneallcroft:
Awesome!!! Thank you so much! : )

allenh1:
(also woot #100 -- that's pretty cool)

PR #101, opened 2017-12-18
Fix an issue with the dockerfile.

allenh1:
This just adds a few commands to the `Dockerfile` that cause problems with (newer) versions of portage.


allenh1:
@tfoote I was kind of expecting that. Fixed.

PR #102, opened 2017-12-19
Add checksum cache

allenh1:
This is an experiment to speed up the generation of OE recipes.

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -84,13 +84,13 @@ def main():
         try:
             md5_file = open('%s/md5_cache.pickle', 'rb')
             md5_cache = pickle.load(md5_file)
-            close(md5_file)
+            md5_file.close()
         except IOError:
             md5_cache = dict()
         try:
             sha256_file = open('%s/md5_cache.pickle', 'rb')
             sha256_cache = pickle.load(sha256_file)
-            close(sha256_file)
+            sha256_file.close()
         except IOError:
             sha256_cache = dict()
     with TempfileManager(args.output_repository_path) as _repo:
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -84,14 +84,14 @@ def main():
         try:
             md5_file = open('%s/md5_cache.pickle', 'rb')
             md5_cache = pickle.load(md5_file)
+            close(md5_file)
         except IOError:
-            md5_file = open('%s/md5_cache.pickle', 'wb')
             md5_cache = dict()
         try:
             sha256_file = open('%s/md5_cache.pickle', 'rb')
             sha256_cache = pickle.load(sha256_file)
+            close(sha256_file)
         except IOError:
-            sha256_file = open('%s/md5_cache.pickle', 'wb')
             sha256_cache = dict()
     with TempfileManager(args.output_repository_path) as _repo:
         if not args.output_repository_path:
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -112,7 +112,9 @@ def main():
                         overlay,
                         regenerate_installer,
                         preserve_existing,
-                        tar_dir
+                        tar_dir,
+                        md5_cache,
+                        sha256_cache
                     )
                 for key in distro_broken.keys():
                     for pkg in distro_broken[key]:
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -80,21 +80,6 @@ def main():
     # open cached tar file if it exists
     md5_cache = None
     sha256_cache = None
-    if args.tar_archive_dir:
-        try:
-            md5_file = open('%s/md5_cache.pickle' % args.tar_archive_dir, 'rb')
-            md5_cache = pickle.load(md5_file)
-            md5_file.close()
-        except IOError:
-            md5_cache = dict()
-        try:
-            sha256_file = open(
-                '%s/sha256_cache.pickle' % args.tar_archive_dir, 'rb'
-            )
-            sha256_cache = pickle.load(sha256_file)
-            sha256_file.close()
-        except IOError:
-            sha256_cache = dict()
     with TempfileManager(args.output_repository_path) as _repo:
         if not args.output_repository_path:
             # give our group write permissions to the temp dir
@@ -107,23 +92,25 @@ def main():
         total_changes = dict()
 
         with TempfileManager(args.tar_archive_dir) as tar_dir:
-            for distro in selected_targets:
-                distro_installers, distro_broken, distro_changes =\
-                    generate_installers(
-                        distro,
-                        overlay,
-                        regenerate_installer,
-                        preserve_existing,
-                        tar_dir,
-                        md5_cache,
-                        sha256_cache
-                    )
-                for key in distro_broken.keys():
-                    for pkg in distro_broken[key]:
-                        total_broken.add(pkg)
+            with HashManager('%s/sha256_cache.pickle') % tar_dir as sha256_cache \
+                 HashManager('%s/md5_cache.pickle') % tar_dir as md5_cache:
+                for distro in selected_targets:
+                    distro_installers, distro_broken, distro_changes =\
+                        generate_installers(
+                            distro,
+                            overlay,
+                            regenerate_installer,
+                            preserve_existing,
+                            tar_dir,
+                            md5_cache,
+                            sha256_cache
+                        )
+                    for key in distro_broken.keys():
+                        for pkg in distro_broken[key]:
+                            total_broken.add(pkg)
 
-                total_changes[distro] = distro_changes
-                total_installers[distro] = distro_installers
+                    total_changes[distro] = distro_changes
+                    total_installers[distro] = distro_installers
 
         num_changes = 0
         for distro_name in total_changes:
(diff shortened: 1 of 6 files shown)


superflore/generators/bitbake/run.py
```diff
@@ -76,6 +77,22 @@ def main():
...
+    # open cached tar file if it exists
+    md5_cache = None
+    sha256_cache = None
+    if args.tar_dir:
+        try:
+            md5_file = open('%s/md5_cache.pickle', 'rb')
+            md5_cache = pickle.load(md5_file)
+            md5_file.close()
+        except IOError:
+            md5_cache = dict()
```
tfoote: Should this be None so it's the same as if the option wasn't passed?

superflore/generators/bitbake/run.py
```diff
@@ -76,6 +77,22 @@ def main():
...
+        try:
+            sha256_file = open('%s/md5_cache.pickle', 'rb')
+            sha256_cache = pickle.load(sha256_file)
+            sha256_file.close()
+        except IOError:
+            sha256_cache = dict()
```
tfoote: Same None here as above?

superflore/generators/bitbake/run.py
```diff
@@ -76,6 +77,22 @@ def main():
...
+    if args.tar_dir:
+        try:
+            md5_file = open('%s/md5_cache.pickle', 'rb')
+            md5_cache = pickle.load(md5_file)
+            md5_file.close()
+        except IOError:
+            md5_cache = dict()
+        try:
+            sha256_file = open('%s/md5_cache.pickle', 'rb')
+            sha256_cache = pickle.load(sha256_file)
```
tfoote: Are there matching write/dump calls to update the cache with the new hashes somewhere?

allenh1: Not yet -- this is still a WIP. Still deciding where to out those.

allenh1:
> Does computing the hashes really become the bottle neck for the builds?

I'm not sure. OE is significantly slower than the ebuild generation, so I'm trying to figure out the source of the slow down. This is really the only extra step that's not in the ebuild generator, so I'm trying to speed things up here. Could be this does nothing.

But this also gives us the ability to remove the tar archives (as only the checksums are needed).

superflore/generators/bitbake/run.py
allenh1: I think my line of thought there was _they want to do a cache, so we'll initialize a dictionary to store that option, so we can check later that it's not `None`._

Sorry, this branch was far from ready for review... I was just opening this so you could see it.

allenh1:
@tfoote Ok, this is the full changeset. Let me know what you think. I'm going to do a benchmark to see if this helps at all.

This holds the full path from the file, including the distro and version.

superflore/generators/bitbake/run.py
```diff
@@ -76,6 +77,24 @@ def main():
         warn('"{0}" distro detected...'.format(args.ros_distro))
         selected_targets = [args.ros_distro]
         preserve_existing = False
+    # open cached tar file if it exists
+    md5_cache = None
+    sha256_cache = None
+    if args.tar_archive_dir:
+        try:
+            md5_file = open('%s/md5_cache.pickle' % args.tar_archive_dir, 'rb')
```
tfoote: I recommend using a context manager here to open and close the file automatically.

https://jeffknupp.com/blog/2016/03/07/python-with-context-managers/

superflore/generators/bitbake/run.py
```diff
@@ -76,6 +77,24 @@ def main():
...
+            md5_cache = pickle.load(md5_file)
+            md5_file.close()
+        except IOError:
+            md5_cache = dict()
+        try:
+            sha256_file = open(
```
tfoote: context manager here too

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -51,11 +54,30 @@ def __init__(self, name, distro, src_uri, tar_dir):
...
-        self.src_sha256 = hashlib.sha256(
-            open(self.getArchiveName(), 'rb').read()).hexdigest()
-        self.src_md5 = hashlib.md5(
-            open(self.getArchiveName(), 'rb').read()).hexdigest()
+        if md5_cache is not None:
+            if self.getArchiveName() not in md5_cache:
+                self.downloadArchive()
+                md5_cache[self.getArchiveName()] = hashlib.md5(
+                    open(self.getArchiveName(), 'rb').read()).hexdigest()
+                md5_file = open('%s/md5_cache.pickle' % tar_dir, 'wb')
```
tfoote: context manager here too

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -51,11 +54,30 @@ def __init__(self, name, distro, src_uri, tar_dir):
...
+            self.src_md5 = md5_cache[self.getArchiveName()]
+        else:
+            self.downloadArchive()
+            self.src_md5 = hashlib.md5(
+                open(self.getArchiveName(), 'rb').read()).hexdigest()
+        if sha256_cache is not None:
+            if self.getArchiveName() not in sha256_cache:
+                sha256_cache[self.getArchiveName()] = hashlib.sha256(
+                    open(self.getArchiveName(), 'rb').read()).hexdigest()
+                sha256_file = open('%s/sha256_cache.pickle' % tar_dir, 'wb')
```
tfoote: context manager here too

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -51,11 +54,30 @@ def __init__(self, name, distro, src_uri, tar_dir):
...
         self.license_md5 = None
         self.tar_dir = tar_dir
-        self.downloadArchive()
-        self.src_sha256 = hashlib.sha256(
-            open(self.getArchiveName(), 'rb').read()).hexdigest()
-        self.src_md5 = hashlib.md5(
-            open(self.getArchiveName(), 'rb').read()).hexdigest()
+        if md5_cache is not None:
+            if self.getArchiveName() not in md5_cache:
+                self.downloadArchive()
```
tfoote: Ahh, this is likely the performance bottleneck. It has to download the archive before computing the hash.

tfoote: Also you download the archive in both the if and else, but then don't download it in the sha256 since it's already there.

Also if you load the cache at the higher level you should write it out again at the higher level instead of each time you process a new package and just update the internal (md5)_cache object for each computation. And write the new cache values out on exit.

It might be worth making a dedicated class that loads holds, and computes the hashes conditionally. It can be parameterized on the hash function. And then you just pass it the md5_cache.get_hash(filename) and it will query if it's stored else compute it. If md5_cache is a context manager, you can pass it the cache filename and hash function. And it will load the file going in and dump the cache on the way out automatically.

allenh1:
Ok, here's the benchmark.

Without cache:
---------------------
 * real: 9 minutes and 3 seconds.
 * user: 17.7 seconds
 * sys: 3.304 seconds

With cache:
-----------------
 * real: 6 minutes and 37 seconds
 * user: 15.5 seconds
 * sys: 2.640 seconds


That's somewhat significant, especially in the real time. This also has the added advantage that we can remove the tars after the pickle cache generates.

superflore/generators/bitbake/yocto_recipe.py
allenh1: > Also if you load the cache at the higher level you should write it out again at the higher level instead of each time you process a new package and just update the internal (md5)_cache object for each computation. And write the new cache values out on exit.

That definitely sounds like a more pythonic way to do this. I'll get right on that.

allenh1:
> Sorry this ended up with a lot of suggestions on a new approach with a HashCache class suggested to allow code reuse and increase readability/decrease branching.

@tfoote Upon implementing, this turned into the `CacheManager` class (which is just a bit more generic).

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -51,11 +53,15 @@ def __init__(self, name, distro, src_uri, tar_dir):
         self.archive_name = None
         self.license_md5 = None
         self.tar_dir = tar_dir
-        self.downloadArchive()
-        self.src_sha256 = hashlib.sha256(
-            open(self.getArchiveName(), 'rb').read()).hexdigest()
-        self.src_md5 = hashlib.md5(
-            open(self.getArchiveName(), 'rb').read()).hexdigest()
+        if self.getArchiveName() not in md5_cache and \
```
tfoote: This should be an `or` in case one or the other is preinitialized but not the other.

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -51,11 +53,15 @@ def __init__(self, name, distro, src_uri, tar_dir):
...
+           self.getArchiveName() not in sha256_cache:
+                self.downloadArchive()
+                md5_cache[self.getArchiveName()] = hashlib.md5(
```
tfoote: This logic could be collapsed into the cache itself if you pass it the hashing function and a lambda or something that would allow it to call self.downloadArchive. But the complexity would be similar so this is fine.

superflore/generators/bitbake/yocto_recipe.py
allenh1: I'll add this in #105 and merge this, since you're mid-travel.

PR #103, opened 2017-12-19
Fix existing version detection.

allenh1:
As pointed out in ros/ros-overlay#279 by @flabrose, the `gpsd_client` packages were out of date.

This would explain why. I'm decently red in the face for not noticing this sooner.


allenh1:
@tfoote sorry -- gotta get another review because of the rebase.

PR #105, opened 2017-12-20
Various Cleanups

allenh1:
This is some cleanup prepping for a new release. I'll update this as I notice things finishing the currently open PR's.


superflore/generators/ebuild/ebuild.py
```diff
@@ -123,7 +123,7 @@ def get_ebuild_text(self, distributor, license_text):
             if len(split) > 1:
                 # they did something like "BSD,GPL,blah"
                 ret += 'LICENSE="( '
-                ret += ' '.join([get_license(l) for l in split])
+                ret += ' '.join([get_license(l.lstrip()) for l in split])
```
tfoote: Why only a lstrip here? You might as well strip both ends of the tokens.

allenh1: That's a good point. Fixed.

PR #106, opened 2017-12-21
Add more oe deps

allenh1:
This adds some deps for Open Embedded, as well as fixes some of the builds.

After this branch is merged, I think we'll do a release. This branch will be ready when catkin is cooperating completely.

### DIFF
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -152,7 +152,9 @@ class yoctoRecipe(object):
         ret += ';md5='
         ret += str(self.license_md5)
         ret += '"\n\n'
-
+        # check for catkin
+        if self.name == 'catkin':
+            ret += 'CATKIN_NO_BIN="True"\n\n'
         # DEPEND
         first = True
         ret += 'DEPENDS = "'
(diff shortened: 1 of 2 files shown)


allenh1:
Ok, there's still some problems. Currently, I'm pretty sure there isn't anything to do in superflore to help with this. I think there's just some tweaking to be done in the `meta-ros` repo.

Newest build failure log is [here](https://gist.github.com/allenh1/907c06ddb709b40270622609402e28d2).

Currently, `ros_comm/rosbag_storage` fails with some macro problem, likely just a missing include folder somewhere. Should be an easy fix.

There's an inability to find `empy` at the moment, despite it being installed in `/usr/lib/python2.7/site-packages`. Since there's no default `empy` installer for OE (in fact the _only_ recipe is located in the `meta-ros` repo), I might just modify the `empy` installer to install in `/opt/ros/[distro]/lib/python2.7/site-packages` like the rest of the ROS packages. @tfoote what do you think about that?

tfoote:
I'd rather see the OE installer for `empy` work in `/usr/lib` as we don't want to start duplicating all system dependencies into `/opt/ros/*` but if that will get things going for now that would be fine. Though if you do that please open a ticket to track pulling it back out. And that would be a good recipie to work toward pushing it to an upstream source like the debian packages we use so that they'll be available more generally.

allenh1:
> I'd rather see the OE installer for empy work in /usr/lib as we don't want to start duplicating all system dependencies into /opt/ros/* but if that will get things going for now that would be fine.

No need! I figured out how to add it to `PYTHONPATH`, and all is now well with detecting `empy`.

PR #108, opened 2017-12-28
Bump version to 0.2.0

allenh1:
Bump version in preparation to release.


allenh1:
@tfoote sorry, I somehow cleared your review while pushing the `CHANGELOG` file. I used [this tool](https://github.com/skywinder/github-changelog-generator) to generate it, and it looks pretty good to me.

allenh1:
@tfoote Pruned. Still a rather large release, though. In the future, I think I'll remember what a patch version is and release them so we don't have these massive releases.

> To release, or not to release.... That is the question...

PR #114, opened 2018-01-07
Add opencv3 check

allenh1:
Resolves ros/ros-overlay#280


allenh1:
@tfoote Struggling to find a good link, but I'll keep looking. [This stackoverflow post](https://stackoverflow.com/questions/6622454/cuda-incompatible-with-my-gcc-version) confirms it, but I'd prefer a more _formal_ source.

Issue #117, opened 2018-01-13
Multi-License Line Fix


allenh1:
This is an easy fix -- just creating an issue for it so I can mark the specific info in the PR to fix it.

https://github.com/ros-simulation/gazebo_ros_pkgs/blob/14db13440b22fada34676ae45b992fd42ec52395/gazebo_ros_pkgs/package.xml#L9

This is not giving us a good license line. Should be the below.
```
LICENSE="( BSD LGPL-2 Apache-2.0 )"
```

PR #118, opened 2018-01-13
Add test for ros-infrastructure/superflore#117.

allenh1:
Add a test to prevent this from happening in the future.

connects to #117.

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -124,7 +124,7 @@ class Ebuild(object):
                 # they did something like "BSD,GPL,blah"
                 ret += 'LICENSE="( '
                 ret += ' '.join([get_license(l.strip()) for l in split])
-                ret += ') "\n'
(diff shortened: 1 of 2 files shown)


allenh1:
@tfoote do you think this merits a patch release? This is not so great.

allenh1:
> The change looks fine. It seems odd to pad whitespace on between the parentheses but if that's their standard, that's what we need to do.

It's not my favorite style either. There's likely some really strange parsing reason for it, but I wouldn't really know.

> I don't know what the ramifications are for this change to judge the need for a patch release.

The only reason I suggest this is so that I don't get any PR's that might have this bug in them. Portage refuses to build an ebuild that is made with that kind of license line (making this quite annoying). I discovered this when I was building `ros-lunar/desktop_full` from scratch on my Gentoo system. All is now working (at least for lunar) with this patch, though.

All of this said, I'm decently sure I'm going to be the only one running superflore for the time being, so I suppose it's not very important that we push a patch version for it.

Issue #116, opened 2018-01-11
Add more information to the generated pull requests


tfoote:
When reviewing the pull requests it would be great to have as much information about where the content came from as possible.

For example this PR: https://github.com/ros/ros-overlay/pull/337 I'd like to know what changes are being integrated into this repo and why?

Adding as much information as possible would be great. In particular things like the command arguments and options used to invoke superflore. As well as any environment and relevant urls for repositories being operated on, the rosdistro etc. Optimally I would be able to completely reproduce the PR with the contents of the description if I wanted to test or try something a little different.

allenh1:
> For example this PR: ros/ros-overlay#337 I'd like to know what changes are being integrated into this repo and why?

That's a good point. Currently, we just display the packages and the distro that's affected. I could see the version changes as a minimal first step for resolving this.

>  In particular things like the command arguments and options used to invoke superflore. As well as any environment and relevant urls for repositories being operated on, the rosdistro etc.

This would also be a super useful feature, and shouldn't be too hard to implement.

Thanks for the feedback, @tfoote.

tfoote:
Looking at https://github.com/ros/ros-overlay/pull/395

```
Superflore ebuild generator ran update from ROS-Overlay commit ee99a55.
To reproduce this PR, run the following command.

/usr/local/bin/superflore-gen-ebuilds
```

One concrete request is to add the superflore version executed and remove the full path (`/usr/local` doesn't make sense on my machine)

allenh1:
> One concrete request is to add the superflore version executed and remove the full path

Ya, that's a good idea. I'm not sure why it expanded, to be honest -- it just pulls from `sys.argv`. I suppose this isn't C, though. Regardless, it's a pretty simple fix. Thanks.

allenh1:
@tfoote Do you think this can be closed? Just cleaning up the backlog.

tfoote:
Before closing this out I'd like to see a little bit more so that the message is understandable for someone who comes across the PR but is not familiar with superflore. There should be enough info for them to find the documentation and read about what's happening so that they could effectively help to review it. Links out to other documentation and an explanation.

PR #119, opened 2018-01-17
Add command line arguments to the PR text.

allenh1:
Work in progress!

connects to #116.


allenh1:
> Right now it's using the defaults, but how would you know what they are from reviewing the PR? Expanding it out into a nice English language sentence would be nice.

That makes a lot of sense. I'll add this in a follow up PR before we close #119.

PR #120, opened 2018-01-19
More Explicit PRs

allenh1:
This changes the style of the ebuild comments to be more human-readable, as suggested by @tfoote.

connects to #119

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -66,8 +66,14 @@ def main():
         nargs='+',
         help='generate only the specified packages'
     )
+    parser.add_argument(
+        '--pr-comment',
+        help='comment to add to the PR',
+        type=str
+    )
     selected_targets = active_distros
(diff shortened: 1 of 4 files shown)


superflore/generators/bitbake/run.py
```diff
@@ -82,6 +88,24 @@ def main():
             os.chmod(_repo, 17407)
         # clone if args.output-repository_path is None
         overlay = RosMeta(_repo, not args.output_repository_path)
+        if not args.only:
+            pr_comment = pr_comment or (
+                'Superflore yocto generator began regeneration of all ' +
+                'packages form ROS distribution(s) %s from allenh1\'s ' +
```
tfoote: Can you make the upstream repo a parameter which defaults to `allenh` instead of having it hard coded?

allenh1: Absolutely. I wanted to add a flag that can redirect the PR in a follow up for both Gentoo and Open Embedded (a `--repository-url` flag).

PR #121, opened 2018-01-20
Record changes for CI

allenh1:
This adds support for CI generation.

### DIFF
--- a/superflore/generators/ebuild/run.py
+++ b/superflore/generators/ebuild/run.py
@@ -85,6 +85,7 @@ def main():
     args = parser.parse_args(sys.argv[1:])
     pr_comment = args.pr_comment
     selected_targets = None
+    preserve_existing = True
     if args.all:
         warn('"All" mode detected... This may take a while!')
         preserve_existing = False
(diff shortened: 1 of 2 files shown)


tfoote:
This is dropping a marker file that says what has changed that you're planning to pick up with the CI testing?

allenh1:
Yes, exactly. It sometimes has enough time to run. Otherwise, I just pull it down and run it locally. I think that'll be the process until we figure out where to spin up the CI.

allenh1:
@tfoote I'm not able to reproduce the test failures here -- not sure what's wrong. Would you mind running the tests on your machine for me?

allenh1:
> This would be good to put into a changelog entry or something like this.

Alright, that makes sense. Maybe we just hold off on this for now? Especially since the success rate is so low.

> But for CI type applications we should find a way to dynamically compute the diff and then operate on that so that if something is rebased the CI will be rerun on the new delta. https://github.com/ros/rosdistro has logic for computing the diff.

That sounds like a better idea. I'm a little embarrassed that had not occurred to me, honestly. And it would be super easy to figure out exactly which files changed by just looking at the file diff (as below).

```
$ git diff --name-only origin/master
```

allenh1:
I'm closing this in favor of a solution that only involves the overlay, and computes the packages to test independent of superflore.

PR #122, opened 2018-01-23
Use allenh1/ros_gentoo_base for Docker

allenh1:
It occurred to me that we really don't need to have two docker images anymore, since the other can just be pulled down when needed.

As such, this simply removes the `repoman_docker` image, as it's a subset of the Dockerfile used to generate `allenh1/ros_gentoo_base`.


tfoote:
It does seem to make sense to collapse those two images together. Might it make sense however to move that image here? Otherwise you'll end up with cross coupling if you want to fork this repo, and then have to fork that image and then redirect to use the alternative version. Versus if it was here it could build it in the case that it's not already hosted.

allenh1:
> Versus if it was here it could build it in the case that it's not already hosted.

That's a good idea. Maybe it should be in its own repo and used as a git submodule? I feel that it doesn't really _belong_ in either repo.

tfoote:
Submodules I've found are often a problem but might work in this case. Though I think that you actually have different use cases. One is the boostrap tool for superflore, and one is a generic base image for users to be able to build on top of. They may be very similar but I don't know that they necessarily will want to maintain all changes on both sides so maintaining them in parallel might make sense.

PS where's the source for that image. I couldn't find it in a quick search.

allenh1:
> Though I think that you actually have different use cases. 

Maybe the solution would then be to host the repoman_docker image?

> PS where's the source for that image. I couldn't find it in a quick search.

You can find the source for it in the [ROS-Overlay repo.](https://github.com/ros/ros-overlay/tree/44992a8ecd8c1dc2dfd0c6779f0fe06b5d994790/docker_file)

PR #125, opened 2018-01-24
Fix CI

allenh1:
@tfoote This is the branch I'll use to fix the CI.

Seems that disabling the pull test does the trick. But the reason I have to is still beyond me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 services: docker
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
-dist: trusty
+dist: xenial
 sudo: required
 language: python
 services: docker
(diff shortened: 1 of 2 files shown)


tfoote:
I'd suggest keeping trying going forward onto xenial. Note that python 3.4 is very old for xenial.

If it works locally, I'd suggest finding a way to disable the pull test only on travis, maybe a command line argument etc.

tfoote:
These issues seem to be in the travis setup before the job even starts. I'd suggest looking through some of the travis tutorials and seeing if there are some basic settings to bring across for docker based testing.

allenh1:
@tfoote I used the `--exclude` flag to skip past the test for `test_pull`. I'm pretty sure it should work now.

allenh1:
Ok, so, some random failures later, and things are ok.

PR #129, opened 2018-02-04
Adding patch file support for Open Embedded.

allenh1:
This includes fies in the `files` directory as patches within the generated recipe text.

connects to #129

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -120,12 +120,10 @@ def main():
         else:
             pr_comment = pr_comment or (
                 'Superflore yocto generator began regeneration of package(s)' +
-                ' %s from ROS distro %s from allenh1 fork of Meta-ROS ' +
-                'from commit %s.' % (
+                ' %s from ROS distro %s from allenh1\'s fork of Meta-ROS ' % (
                     args.only,
-                    args.ros_distro,
-                    overlay.repo.get_last_hash()
-                )
+                    args.ros_distro
+                ) + ' from commit %s.' % (overlay.repo.get_last_hash())
             )
         # generate installers
         total_installers = dict()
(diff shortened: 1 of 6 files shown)


allenh1:
current result is allenh1/meta-ros#28

superflore/generators/bitbake/run.py
```diff
@@ -99,10 +99,11 @@ def main():
             )
         else:
             pr_comment = pr_comment or (
-                'Superflore yocto generator began regeneration of package(s)' +
-                ' %s from ROS distro %s from allenh1\'s fork of Meta-ROS ' +
-                'from commit %s.' % (
+                'Superflore yocto generator began regeneration of package(s)'
+                ' %s from ROS distro %s from allenh1\'s fork of Meta-ROS'
```
tfoote: This is missing a parameterization based on #128

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -173,5 +174,17 @@ def get_recipe_text(self, distributor, license_text):
         ret += 'SRC_URI[sha256sum] = "' + self.src_sha256 + '"\n'
         ret += 'S = "${WORKDIR}/'
         ret += self.get_src_location() + '"\n\n'
+        # Check for patches
+        if self.patch_files:
+            ret += 'SRC_URI += "\\\n'
+            ret += ' \\\n'.join(self.patch_files) + '"\n\n'
+        if self.name == 'catkin':
```
tfoote: Can these be integrated into patches?

allenh1: In the normal sense, no. I think I could, however, just detect an existing `.inc` file and include it.

superflore/generators/bitbake/run.py
allenh1: Honestly, I think the `allenh1` part can be removed.

Issue #130, opened 2018-02-04
Add `--dry-run` and `--pr-only` to Open Embedded


allenh1:
Open Embedded is lagging behind in a few places, one of which is the `--pr-only` and `--dry-run` flags that can be passed to the ebuild generator to prevent an unexpected PR from filing.

Should just be a replication of logic.

connects to #49.

allenh1:
Resolved by #131

PR #131, opened 2018-02-06
Add --dry-run and --pr-only to Open Embedded

allenh1:
Work in progress!

So far, only the save and load logic has been centralized. Opening for visibility.

connects to #130.


allenh1:
@tfoote I think this one is ready now. The only really questionable part of this PR is the centralization of the parser logic. The only reason for this is to reduce the code duplication among generators, and make it so that the arguments don't have to be listed every time. As this was just an experiment, we can, if this is not your favorite idea, I made it so I can revert it pretty easily if you object. Just let me know.

Issue #67, opened 2017-08-25
2FA breaks everything


allenh1:
After updating my GitHib account to two-factor authentication, Superflore couldn't file pull requests in the usual fashion.

tfoote:
You need to setup an oauth token and use it in the same way as bloom: http://wiki.ros.org/bloom/Tutorials/GithubManualAuthorization

allenh1:
Temporary Fix (before I actually integrate the OAth stuff):

Make an oath token and replace the password in the `.netrc` file with the token you get back.

PR #132, opened 2018-02-06
Use GitHub api instead of the Git Pull Request module

allenh1:
Use the GitHub API to file PRs, removing the `git-pull-request` dependency.

connects to #67


allenh1:
@tfoote Ready for review

allenh1:
That sounds good to me. Will add that in a follow up.

PR #133, opened 2018-02-08
Add stdeb.cfg

allenh1:
I couldn't re-open the other PR, but could (somehow) restore the deleted branch.

This is updated and has been rebased on top of the current master. I'm going to try it out in a container.

connects to #78

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,13 +15,13 @@ install_requires = [
     'docker',
     'pyyaml',
     'pygithub',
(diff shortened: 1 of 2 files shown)


allenh1:
@tfoote / @nuclearsandwich I actually just have no idea what I'm doing, so I've uploaded the error log [here](https://gist.github.com/allenh1/4b777acef153f0121dccf5f64c00d81b). Let me know if you have any suggestions.

PR #134, opened 2018-02-08
Centralize Duplicate Code

allenh1:
This helps de-duplicate a lot of the parsing code. Just makes it a little more convenient to have a namespace interface.


nuclearsandwich:
Is there a reason for superflore to use its own package manifest parser rather than relying on catkin_pkg?

allenh1:
@nuclearsandwich That's a good question -- I'll check that. Would be nice to cut out this section of code.

nuclearsandwich:
> That's a good question -- I'll check that. Would be nice to cut out this section of code.

Here's an example of bloom's entrypoint into package manifest parsing. You can point catkin_pkg at a directory and it will find packages whether you give it a repository root with multiple packages in subdirectories or a single package.

https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/generate_cmd.py#L102

allenh1:
@nuclearsandwich I centralized the `catkin_pkg` logic, thanks for pointing me to it!

allenh1:
Ok, I think I'm done with this one now. Using some tools (namely `pylint`), I found some duplicated sections of the code and did some significant reduction.

nuclearsandwich:
Make sure you add catkin_pkg to the setup.py. I'd suggest requiring at least 0.4.0 as that is the version that includes package.xml format 3 support.

allenh1:
I should actually add some tests for the new functions first.

superflore/utils.py
```diff
@@ -235,3 +235,38 @@ def _resolve_dep_open_embedded(pkg):
...
+        for pkg in sorted(missing_list):
+            missing_deps += " * [ ] {0}\n".format(pkg)
+    return missing_deps or 'No missing dependencies.\n'
+
+
+def github_url_to_repo_org(url):
+    """Extract owner and repository from GitHub url."""
+    # check that the upstream_repo is a github repo
+    if 'github.com' not in url:
+        raise RuntimeError('Non-GitHub repos are not supported!')
```
tfoote: For someone reading this exception on it's own, say what is not supported. (aka extracting github org and repo name from expected github url)

tfoote: In the future this likely could be extended to support other hosting providers that also have APIs. (aka bitbucket and gitlab) It might make sense to not embed github in the function name only in the error message.

allenh1: That makes sense. Will do.

allenh1:
@tfoote Ok, this is ready for another look now.

PR #135, opened 2018-02-14
Be My Ardent-tine

allenh1:
This is a work in progress still! I'm not super pleased with the quality of this patch.

But, in its current state, this _does_ add ROS 2 support, which is pretty neat.

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -121,7 +121,7 @@ class Ebuild(object):
 
     def get_inherit_line(self):
         # if we are using catkin, we just inherit ros-cmake
-        if self.build_type == 'catkin':
+        if self.build_type in ['catkin', 'cmake']:
             return 'inherit ros-cmake\n\n'
         elif self.build_type == 'ament_python':
             return 'inherit ament-python\n\n'
@@ -231,12 +231,13 @@ class Ebuild(object):
         ret += "ROS_PREFIX=\"opt/ros/${ROS_DISTRO}\"\n"
         # Patch source if needed.
         if self.has_patches:
+            # TODO(allenh1): explicitly list patches
             ret += "\nsrc_prepare() {\n"
             ret += "    cd ${P}\n"
             ret += "    EPATCH_SOURCE=\"${FILESDIR}\""
             ret += " EPATCH_SUFFIX=\"patch\" \\\n"
             ret += "    EPATCH_FORCE=\"yes\" epatch\n"
-            if self.build_type == 'catkin':
+            if self.build_type in ['catkin', 'cmake']:
                 ret += "    ros-cmake_src_prepare\n"
             ret += "}\n"
         # source configuration
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -107,8 +107,8 @@ class Ebuild(object):
         ret += " license\n\n"
         return ret
 
-    def get_eapi(self):
-        return 'EAPI=%d\n' % self.eapi
+    def get_eapi_line(self):
+        return 'EAPI=%s\n' % self.eapi
 
     def get_python_compat(self, python_versions):
         ver_string = ''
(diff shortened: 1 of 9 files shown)


allenh1:
Next up is to add support for things like python (for using distutils instead of CMake), as, currently, ROS2 ebuilds fail due to a lack of `CMakeLists.txt` file.

Also, there's complaints about a missing opensplice dependency, so nothing is buildable past the type support.

superflore/generators/ebuild/overlay_instance.py
```diff
@@ -40,6 +40,7 @@ def commit_changes(self, distro):
             'lunar': 'regenerate ros-lunar, ',
             'indigo': 'regenerate ros-indigo, ',
             'kinetic': 'regenerate ros-kinetic, ',
+            'ardent': 'regenerage ros2-ardent, ',
```
tfoote: typo 'regenerage'

nuclearsandwich:
> Next up is to add support for things like python (for using distutils instead of CMake), as, currently, ROS2 ebuilds fail due to a lack of CMakeLists.txt file.

It would be a big change, and maybe not worth making at the moment, but adding ebuild generation to bloom and then using it in superflore might make sense long term.

Since 0.6.0 bloom uses different templates for generating different package build types. Even if it doesn't make sense to try and incorporate bloom now, you can still see how the debian packages are constructed. https://github.com/ros-infrastructure/bloom/tree/master/bloom/generators/debian/templates

allenh1:
Ok, I'm a little upset about these random failures on Travis. Under "pr", the CI passes for pypy3, and for "push" it fails only for python 3.5. And for string orderings? I don't get it.

Must be some `LC_LANG` setting or something that's disrupting the sort function.

superflore/generators/ebuild/ebuild.py
```diff
@@ -225,19 +225,28 @@ def get_ebuild_text(self, distributor, license_text):
         # CMAKE_BUILD_TYPE
         if self.name == "catkin":
             ret += "BUILD_BINARY=\"0\"\n"
+
         ret += "ROS_DISTRO=\"{0}\"\n".format(self.distro)
         ret += "ROS_PREFIX=\"opt/ros/${ROS_DISTRO}\"\n"
-
+        # TODO(allenh1): This is gross. No.
```
nuclearsandwich: I'm regularly guilty of this as well, but instead of saying "This is gross." A more useful XXX or TODO comment might give an example of when the code will fail

allenh1: It's just there so I fix it before this gets merged (I will not merge that into the codebase -- that's just awful).

allenh1: Actually, on second thought, I really don't think this is as awful as I initially thought it was.

The more robust solution is to create an `eclass` for this, but that's not really in the spirit of things if it's just for unpacking things. So I think I'll just leave a note that says "if you need to update this, then do so. Otherwise, this is sufficient."

allenh1:
I think this just about does it for this PR. I ended up adding in a template `ament-python.eclass`. I think there should be a follow up with some tests for the new logic introduced here, but I don't want to add that to this PR (it's already pretty bloated).

tl;dr The remaining fixes will be done to the eclasses, so this can be merged.

Next step there is getting @nuclearsandwich's `ros_workspace` to build. Currently, it's [not](https://gist.github.com/allenh1/48e80007fc07efafccd9561bdaded8c9) (likely due to a multilib mapping issue).

superflore/generators/ebuild/run.py
```diff
@@ -31,12 +32,16 @@
...
 from superflore.utils import load_pr
 from superflore.utils import ok
+from superflore.utils import ros2_distros
 from superflore.utils import save_pr
 from superflore.utils import url_to_repo_org
 from superflore.utils import warn
 
-# Modify if a new distro is added
-active_distros = ['indigo', 'kinetic', 'lunar']
+# TODO(allenh1): It would be super nice make this a configuration option.
```
tfoote: :+1:  `ROSDISTRO_INDEX_URL`

Just use `rosdistro.get_index_url()` https://github.com/ros-infrastructure/rosdistro/blob/master/src/rosdistro/__init__.py#L68

allenh1: I figured there was going to be something like that. I'll look into this in a follow up. Thanks.

PR #157, opened 2018-07-05
Support python3_6 for ROS 1

concavegit:
This addresses https://github.com/ros/ros-overlay/issues/642


allenh1:
@concavegit I seem to have made a mistake in assisting you! There's a couple more places to add it.

The example ebuild used for tests:

https://github.com/ros-infrastructure/superflore/blob/02b0dcddc7e954713952e7c368d254ae4aa55aca/tests/ebuild/simple_expected.ebuild#L5

Test for the python version:

https://github.com/ros-infrastructure/superflore/blob/02b0dcddc7e954713952e7c368d254ae4aa55aca/tests/test_ebuild.py#L156

After those two changes are done, this looks good to me.

Issue #163, opened 2019-01-12
Bad license with "3-Clause BSD"


awesomebytes:
```
superflore-gen-ebuilds --ros-distro kinetic --only move_base_flex mbf_utility mbf_simple_nav mbf_msgs mbf_costmap_nav mbf_costmap_core mbf_abstract_nav mbf_abstract_core
>>>> Working in temporary directory /tmp/tmpxs1__7ca
>>>> Creating new branch gentoo-bot-ufXOtSTWdJ...
>>>> 
>>>> Regenerating package 'move_base_flex'...
>>>> Successfully generated installer for package 'move_base_flex'.
>>>> Regenerating package 'mbf_utility'...
>>>> Successfully generated installer for package 'mbf_utility'.
>>>> Regenerating package 'mbf_simple_nav'...
!!!! Could not match license "3-Clause BSD".
>>>> Cleaning up temporary directory /tmp/tmpxs1__7ca
Traceback (most recent call last):
  File "/home/sam/.local/bin/superflore-gen-ebuilds", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.5/dist-packages/superflore-0.2.1-py3.5.egg/superflore/generators/ebuild/run.py", line 124, in main
  File "/usr/local/lib/python3.5/dist-packages/superflore-0.2.1-py3.5.egg/superflore/generators/ebuild/gen_packages.py", line 85, in regenerate_pkg
  File "/usr/local/lib/python3.5/dist-packages/superflore-0.2.1-py3.5.egg/superflore/generators/ebuild/gen_packages.py", line 216, in ebuild_text
  File "/usr/local/lib/python3.5/dist-packages/superflore-0.2.1-py3.5.egg/superflore/generators/ebuild/ebuild.py", line 174, in get_ebuild_text
  File "/usr/local/lib/python3.5/dist-packages/superflore-0.2.1-py3.5.egg/superflore/utils.py", line 210, in get_license
superflore.exceptions.UnknownLicense: bad license

```

That's what I got. I'm looking around to see how to overcome it, but maybe you'll be faster to spot what to do.

PR #164, opened 2019-01-12
Support BSD 'X-Clause BSD' license string

awesomebytes:
As mentioned in https://github.com/ros-infrastructure/superflore/issues/163

### DIFF
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -175,7 +175,7 @@ def get_license(l):
         if not version:
             version = matches.group(5)
         if version:
-            return 'BSD-{0}'.format(version) 
+            return 'BSD-{0}'.format(version)
         return 'BSD'
     elif re.search(lgpl_re, l, f):
         version = re.search(lgpl_re, l, f)
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -148,7 +148,7 @@ def trim_string(string, length=80):
 
 
 def get_license(l):
-    bsd_re = '^(BSD)((.)*([124]))?'
+    bsd_re = '^(BSD)((.)*([124]))?|([1234])(.)?Clause(.)*(BSD)'
     gpl_re = '((([^L])*(GPL)([^0-9]*))|'\
         '(GNU(.)*GENERAL(.)*PUBLIC(.)*LICENSE([^0-9])*))([0-9])?'
     lgpl_re = '(((LGPL)([^0-9]*))|'\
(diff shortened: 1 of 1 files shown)


superflore/utils.py
```diff
@@ -148,7 +148,7 @@ def trim_string(string, length=80):
 
 
 def get_license(l):
-    bsd_re = '^(BSD)((.)*([124]))?'
+    bsd_re = '^(BSD)((.)*([124]))?|([1234])(.)?Clause(.)*(BSD)'
```
awesomebytes: @allenh1 My regex skills are pretty dusty, I tried to make it as safe as possible... so I just added a second option to match, the original one is untouched.

superflore/utils.py
```diff
@@ -170,9 +170,12 @@ def get_license(l):
             return 'Apache-%.1f' % (float(version))
         return 'Apache-1.0'
     elif re.search(bsd_re, l, f):
-        version = re.search(bsd_re, l, f).group(4)
+        matches = re.search(bsd_re, l, f)
+        version = matches.group(4)
+        if not version:
+            version = matches.group(5)
```
awesomebytes: @allenh1  And here I just try to get the version from either the previous kind of match, or the new (giving priority to the previous case).

allenh1:
```
superflore/utils.py:178:45: W291 trailing whitespace\n1\n'
```

The linter is angry.

allenh1:
Other than the linter failures, I would be really grateful if you could add some tests for the license check. You should be able to find some here.

https://github.com/ros-infrastructure/superflore/blob/adf2fc5d9cccc2fdb92a653718e96b74534b6b74/tests/test_utils.py#L82-L123

This just helps prevent this regex getting broken in the future.

awesomebytes:
> ```
> superflore/utils.py:178:45: W291 trailing whitespace\n1\n'
> ```
> The linter is angry.

Fixed

allenh1:
@awesomebytes did you mean to close this?

awesomebytes:
I messed up the PR. And I messed up my Python installation so I can't run superflore (neither the tests) anymore.

If you wanna give it a try to fix it... you are welcome, otherwise I'll need to revisit this some other day.

Note that I just noticed a logic error. If `'3-Clause BSD'` is the license string, by your previous code you return `BSD` but I was returning `BSD-3`. So there must be another `if matches.group(5) == 3: return 'BSD'` or something like that.

Sorry for the mess. I should go to sleep.

allenh1:
@awesomebytes no worries -- we can just fix it at some other point.

PR #168, opened 2019-03-30
Initial OpenEmbedded Support for ROS 2

andre-rosa:
Initial OpenEmbedded support for superflore

  * Requires related changes in rospkg, rosdep and rosdistro
     * [x] https://github.com/ros-infrastructure/rospkg/pull/166
     * [x] https://github.com/ros-infrastructure/rosdep/pull/673
     * [x] https://github.com/ros/rosdistro/pull/20763
  * Able to successfully replace the manually crafted Bouncy ROS 2 recipes in 
    https://github.com/lgsvl/build-ros2-lgsvl by the autogenerated ones, then build and run.
  * Able to successfully build core Crystal ROS 2 image. Complete Crystal build is an ongoing effort.
  * Please see the proposal below:
    https://discourse.ros.org/t/a-proposal-for-a-superflore-oe-recipe-generation-scheme/8401

Tagging @rojkov @bulwahn @johannesschrimpf @mpthompson @dkargin @dominiquehunziker @mdhorn << Could you please take a look?

PR #169, opened 2019-04-02
Small cleanups

andre-rosa:
Separating few cleanups from https://github.com/ros-infrastructure/superflore/pull/168 .

### DIFF
--- a/superflore/generators/bitbake/gen_packages.py
+++ b/superflore/generators/bitbake/gen_packages.py
@@ -71,7 +71,7 @@ def regenerate_installer(
             for f in glob.glob('%s*.patch' % patch_path)
         ]
     if preserve_existing and existing:
-        ok("recipe for package '%s' up to date, skpping..." % pkg)
+        ok("recipe for package '%s' up to date, skipping..." % pkg)
         return None, []
     elif existing:
         overlay.repo.remove_file(existing[0])
--- a/superflore/generators/bitbake/gen_packages.py
+++ b/superflore/generators/bitbake/gen_packages.py
@@ -74,7 +74,7 @@ def regenerate_installer(
         ok("recipe for package '%s' up to date, skipping..." % pkg)
         return None, []
     elif existing:
-        overlay.repo.remove_file(existing[0])
+        overlay.repo.remove_file(existing[0], True)
     try:
         current = oe_installer(
             distro, pkg, tar_dir, md5_cache, sha256_cache, patches, inc_files
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -23,6 +23,7 @@ from superflore.generators.bitbake.ros_meta import RosMeta
 from superflore.parser import get_parser
 from superflore.repo_instance import RepoInstance
 from superflore.TempfileManager import TempfileManager
+from superflore.utils import active_distros
 from superflore.utils import clean_up
 from superflore.utils import err
 from superflore.utils import file_pr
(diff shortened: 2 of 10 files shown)


superflore/utils.py
```diff
@@ -202,7 +199,7 @@ def get_license(l):
     elif re.search(cc_re, l, f):
         return 'CC-BY-SA-3.0'
     elif re.search(boost_re, l, f):
-        return 'Boost-1.0'
```
allenh1: Please revert this change

allenh1: This still does not work with Gentoo, which gets very upset if this is not matched exactly.

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -58,17 +58,14 @@ def __init__(
...
-                self.downloadArchive()
-                md5_cache[self.getArchiveName()] = hashlib.md5(
-                    open(self.getArchiveName(), 'rb').read()).hexdigest()
-                sha256_cache[self.getArchiveName()] = hashlib.sha256(
-                    open(self.getArchiveName(), 'rb').read()).hexdigest()
+            self.downloadArchive()
+            md5_cache[self.getArchiveName()] = hashlib.md5(
+              open(self.getArchiveName(), 'rb').read()).hexdigest()
+            sha256_cache[self.getArchiveName()] = hashlib.sha256(
+              open(self.getArchiveName(), 'rb').read()).hexdigest()
```
allenh1: I think the linter is unhappy with this whitespace change.

superflore/generators/bitbake/run.py
```diff
@@ -31,38 +32,29 @@
...
 from superflore.utils import warn
 
-# Modify if a new distro is added
-active_distros = ['indigo', 'kinetic', 'lunar']
-
 
 def main():
-    preserve_existing = True
     overlay = None
+    preserve_existing = True
```
allenh1: to reduce the delta, does the location of this variable need to change?

superflore/generators/bitbake/gen_packages.py
```diff
@@ -71,10 +71,10 @@ def regenerate_installer(
             for f in glob.glob('%s*.patch' % patch_path)
         ]
     if preserve_existing and existing:
-        ok("recipe for package '%s' up to date, skpping..." % pkg)
+        ok("recipe for package '%s' up to date, skipping..." % pkg)
```
allenh1: *facepalm* and to think, I almozt niver hab tipos

.vscode/launch.json
```diff
@@ -0,0 +1,57 @@
+{
```
allenh1: I'm not sure how I feel about the `.vscode` files being added, to be frank. I don't hate this in any sense, but, as I don't use this editor (if you can't tell from my atrocious spelling abilities), I'm not sure I want to be responsible for maintaining this.

Is this something that should absolutely be included in the repository, or could you simply hold on to these files?

@tfoote do you have any opinions on this? I'm not just not sure this is best practice.

superflore/utils.py
andre-rosa: Done; bitbake is also happy about 'Boost-1.0'.

superflore/generators/bitbake/run.py
andre-rosa: Right, otherwise it's a small one line delta.

.vscode/launch.json
andre-rosa: I can definitely remove it. I thought that would be helpful for other people and non-intrusive but it's totally fine to carry it myself.

tfoote: Yeah, in general we try to avoid development tool specific as it's both a personal choice of what editor to use as well as how you want to set it up.

andre-rosa: Thanks, I've dropped the .vscode files.

allenh1:
lgtm -- waiting for @tfoote to review and we should be ready to merge this one.

Thanks!

andre-rosa:
Thank you _both_!

PR #171, opened 2019-04-09
Use index for querying distro information

andre-rosa:
* Today Mon. Apr 8th we got 'dashing' alpha release
    https://index.ros.org//doc/ros2/Releases/Release-Dashing-Diademata/

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -23,16 +23,15 @@ from superflore.generators.bitbake.ros_meta import RosMeta
 from superflore.parser import get_parser
 from superflore.repo_instance import RepoInstance
 from superflore.TempfileManager import TempfileManager
-from superflore.utils import active_distros
 from superflore.utils import clean_up
 from superflore.utils import err
 from superflore.utils import file_pr
 from superflore.utils import gen_delta_msg
 from superflore.utils import gen_missing_deps_msg
+from superflore.utils import get_distros_by_status
 from superflore.utils import info
 from superflore.utils import load_pr
 from superflore.utils import ok
-from superflore.utils import ros2_distros
 from superflore.utils import save_pr
 from superflore.utils import url_to_repo_org
 from superflore.utils import warn
@@ -73,7 +72,7 @@ def main():
         selected_targets = [args.ros_distro]
         preserve_existing = False
     if not selected_targets:
-        selected_targets = active_distros + ros2_distros
+        selected_targets = get_distros_by_status('active')
     repo_org = 'allenh1'
     repo_name = 'meta-ros'
     if args.upstream_repo:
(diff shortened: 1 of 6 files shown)


allenh1:
Lgtm. Leaving for @tfoote to merge

nuclearsandwich:
Now that the index-v4.yaml file includes information about ROS 1 vs ROS 2 distributions. Is it possible to use the index rather than hard-coding distribution lists in the tool?

nuclearsandwich:
In fact, looking more closely at [the source](https://github.com/ros-infrastructure/superflore/blob/abb1352ab518dd442cfd8d1ded747a7c6b79fe8c/superflore/generators/bitbake/ros_meta.py#L43-L53) I'm not sure that the commit message formatting even merits working that hard. And it makes it harder for orgs with forked rosdistros to use superflore.

andre-rosa:
> I've opened #172 which removes the hard coded distro lists from commit messages but couldn't quickly replace the last hard coded list. I still believe it's possible to do so and have left details on the other PR.

Great @nuclearsandwich @allenh1 , thanks! I pushed a change on top of yours adding the support to get the hardcoded information from the index instead. Could you please take a look?

andre-rosa:
@tfoote << Thank you for reviewing! I rebased it on top of the dependent changes. Could you please take another look?

superflore/generators/bitbake/run.py
```diff
@@ -73,7 +72,7 @@ def main():
         selected_targets = [args.ros_distro]
         preserve_existing = False
     if not selected_targets:
-        selected_targets = active_distros + ros2_distros
+        selected_targets = get_distros_by_status('active')
```
nuclearsandwich: Just a note, this will not get `dashing` at the moment as it's currently in `prerelease`.

superflore/utils.py
```diff
@@ -219,6 +213,18 @@ def resolve_dep(pkg, os):
...
+def get_distros():
+    index = get_cached_index()
+    return index.distributions
+
+
+def get_distros_by_status(status='active'):
+    distros = get_distros()
+    distros_by_status = [d for d in distros.keys(
+    ) if distros[d].get('distribution_status') == status]
+    return distros_by_status
```
nuclearsandwich: If superflore is python3 only this could be simplified to

```python
    return [d for d, info in get_distros().items()
            if info['distribution_status'] == status]

superflore/generators/bitbake/run.py
andre-rosa: I believe that's okay as its [distribution](https://raw.githubusercontent.com/ros/rosdistro/master/dashing/distribution.yaml) is not complete anyway. We still can specify it by using ```--ros-distro```.

superflore/utils.py
andre-rosa: Done, thanks!

andre-rosa:
Thanks, I'll rebase now!

PR #173, opened 2019-04-09
Add distro parameter to resolve_dep

andre-rosa:
@allenh1  << Are you able to verify that passing the distro parameter to resolve_rosdep_key in 'gentoo' case works well too?


andre-rosa:
Thanks, I'll rebase now!

Issue #170, opened 2019-04-03
Centralize skip keys option for dropping dependencies


andre-rosa:
Concept is similar to [rosdep's skip-keys](http://docs.ros.org/independent/api/rosdep/html/commands.html), where we want to explicitly skip some packages and not adding them as dependencies.

Please see the [related change](https://github.com/ros-infrastructure/superflore/pull/168#discussion_r271119161) for bitbake.

PR #175, opened 2019-04-16
Add `--skip-keys` option

andre-rosa:


superflore/generators/bitbake/gen_packages.py
```diff
@@ -35,8 +35,12 @@
 
 
 def regenerate_installer(
-    overlay, pkg, distro, preserve_existing, tar_dir, md5_cache, sha256_cache
+    overlay, pkg, distro, preserve_existing, tar_dir, md5_cache, sha256_cache,
+    skip_keys
 ):
+    if pkg in skip_keys:
+        warn("Package '%s' is on skip-keys, skipping..." % pkg)
```
allenh1: I'm not a huge fan of the phrasing here. Maybe something like `"Package '%s' is in skip-keys list, skipping..."`? I'm not sure.

superflore/generators/ebuild/gen_packages.py
```diff
@@ -42,7 +42,10 @@
 org_license = "BSD"
 
 
-def regenerate_pkg(overlay, pkg, distro, preserve_existing=False):
+def regenerate_pkg(overlay, pkg, distro, preserve_existing, skip_keys):
```
allenh1: I don't remember what the implication of `preserve_existing` not defaulting to `False` is, but would probably feel better if it continued to default to `False`.

superflore/parser.py
```diff
@@ -59,4 +59,10 @@ def get_parser(tool_tip, is_generator=True):
             help='location of the upstream repository',
             type=str
         )
+        parser.add_argument(
+            '--skip-keys',
+            nargs='+',
+            help='The specified rosdep keys will be ignored. '
```
allenh1: I don't think "rosdep keys" is the correct term. I think this should read `'packages to skip during regeneration'`. If you intend to skip any packages depending on a certain `rosdep` key, I think that is already handled (at least it is on Gentoo).

superflore/generators/bitbake/gen_packages.py
andre-rosa: Done, thanks.

superflore/generators/ebuild/gen_packages.py
andre-rosa: No worries; done.

superflore/parser.py
andre-rosa: Done

superflore/generators/bitbake/gen_packages.py
```diff
@@ -35,8 +35,12 @@
 
 
 def regenerate_installer(
-    overlay, pkg, distro, preserve_existing, tar_dir, md5_cache, sha256_cache
+    overlay, pkg, distro, preserve_existing, tar_dir, md5_cache, sha256_cache,
+    skip_keys=None
 ):
+    if skip_keys and pkg in skip_keys:
```
allenh1: Sorry, I didn't realize this until just now... Can we handle this `if` statement in the caller of this function? I think that would make it harder to forget on any platforms that get added in the future.

andre-rosa: Sure; could you please take a look?

allenh1:
@tfoote I'm going to run a quick test locally, then I'll merge this.

andre-rosa:
Thank you for reviewing! I'll rebase the other changes on top.

andre-rosa:
Closes: #170

PR #176, opened 2019-04-18
Pass distro to generate_installer

andre-rosa:
* OpenEmbedded generation will need information from distro from
    run.py, so allow run.py to get_distro() just once and pass it to
    generate_installer(), like it's done for regenerate_installer().

### DIFF
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from rosinstall_generator.distro import get_distro
 from rosinstall_generator.distro import get_package_names
 from superflore.exceptions import UnknownBuildType
 from superflore.exceptions import UnknownLicense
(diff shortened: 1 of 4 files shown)


superflore/generate_installers.py
```diff
@@ -24,14 +23,14 @@
...
-    distro_name,             # ros distro name
+    distro,                  # ros distro
     overlay,                 # repo instance
     gen_pkg_func,            # function to call for generating
     preserve_existing=True,  # don't regenerate if installer exists
     *args,                   # any additional args for gen_pkg_func
     **kwargs                 # any additional keyword arguments
 ):
-    distro = get_distro(distro_name)
+    distro_name = distro._distribution_file.name
```
tfoote: It's better not to use the "private" attribute prefixed by underscore here. You can just use distro.name. 

All the elements are proxied through: https://github.com/ros-infrastructure/rosdistro/blob/e03a8687d7d4253fad08e4ff80e7c0f5c8fb1984/src/rosdistro/distribution.py#L62

```suggestion
    distro_name = distro.name
```

andre-rosa: Great, thank you!

PR #177, opened 2019-04-18
Rename regenerate_installer to regenerate_pkg

andre-rosa:
* Like in ebuild, for consistency's sake.


allenh1:
Good call. I'll just merge this since CI is green.

PR #179, opened 2019-04-24
Inform that --only option also requires setting --ros-distro

andre-rosa:


andre-rosa:
> lgtm
> 
> thanks for the patch!

Thank _you_ @allenh1!

PR #180, opened 2019-04-24
Add `--upstream-branch` option

andre-rosa:

PR #182, opened 2019-04-25
Fix args.all handling

andre-rosa:
- distro could never be 'all' after all

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -188,7 +188,7 @@ def main():
         delta = gen_delta_msg(total_changes)
         missing_deps = gen_missing_deps_msg(total_broken)
         # Commit changes and file pull request
-        overlay.commit_changes(args.ros_distro)
+        overlay.commit_changes('all' if args.all else args.ros_distro)
         if args.dry_run:
             info('Running in dry mode, not filing PR')
             save_pr(
--- a/superflore/generators/bitbake/ros_meta.py
+++ b/superflore/generators/bitbake/ros_meta.py
@@ -36,16 +36,12 @@ class RosMeta(object):
 
     def commit_changes(self, distro):
         info('Adding changes...')
-        if distro == 'all' or distro == 'update':
+        if distro == 'all':
+            commit_msg = 'regenerate all distros, {0}'
             self.repo.git.add('recipes-ros-*')
         else:
-            self.repo.git.add('recipes-ros-{0}'.format(distro))
-        if distro == 'update':
-            commit_msg = 'rosdistro sync, {0}'
-        elif distro == 'all':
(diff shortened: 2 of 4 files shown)


andre-rosa:
@allenh1 ptal

tests/test_utils.py
```diff
@@ -64,7 +64,7 @@ def test_trim_string(self):
         self.assertEqual(ret, 'a[...]')
 
     def test_mkdir(self):
-        """Tests the make directory funciton"""
+        """Tests the make directory function"""
```
allenh1: I prefer the spelling fixes be handled in their own PR if possible

superflore/generators/bitbake/run.py
```diff
@@ -188,7 +189,7 @@ def main():
         delta = gen_delta_msg(total_changes)
         missing_deps = gen_missing_deps_msg(total_broken)
         # Commit changes and file pull request
-        overlay.commit_changes(args.ros_distro)
+        overlay.commit_changes('all' if args.all else args.ros_distro)
```
allenh1: I think this is the only change in this PR that is fixing an issue? If you could please file an issue for that problem as well that would be really great (for tracking/changelog generation for the next release).

andre-rosa:
> Seems there's a lot of extra stuff in this PR -- If it does not depend on #180, I would strongly prefer you reduce it to only the issue it is addressing so we can get it out of the way of other contributions.
> 
> Thanks for the fix!

Thanks, yes; it is just rebased on top of other PRs. ;)

CHANGELOG.md
```diff
@@ -30,11 +30,11 @@
 **Implemented enhancements:**
 
 - Move some packages to DEPEND [\#83](https://github.com/ros-infrastructure/superflore/issues/83)
-- Create a `tar-archive-dir` to clean up the `.tar` files downloaded for Open Embedded [\#80](https://github.com/ros-infrastructure/superflore/issues/80)
```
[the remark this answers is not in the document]
andre-rosa: Updated in #181.

superflore/generators/bitbake/run.py
andre-rosa: Yes, thanks! Fixes #183

allenh1:
I'll check this one & then merge it

allenh1:
Filed ros/ros-overlay#796 with this in place -- worked like a charm! (there were not any updates, so I just re-generated dashing).

Thanks for all the patches! We're getting closer!

andre-rosa:
> Filed [ros/ros-overlay#796](https://github.com/ros/ros-overlay/pull/796) with this in place -- worked like a charm! (there were not any updates, so I just re-generated dashing).
> 
> Thanks for all the patches! We're getting closer!

Thanks a bunch, @allenh1!

PR #185, opened 2019-05-02
Pass along unknown licenses unclassified

### LINKED ISSUE #184: Melodic qt_metapackages cannot be generated
Currently, it fails with ```!!!! Could not match license "United States Government Purpose".``` and throws ```UnknownLicense: bad license```

andre-rosa:
- Fixes #184 generating qt_metapackages recipes in Melodic
- Currently Proprietary and United States Government Purpose licenses
  aren't matched in SuperFlore and it aborts generation of some melodic
  packages like qt_metapackages/libqt_core
  https://github.com/swri-robotics/qt_metapackages/blob/master/libqt_core/package.xml
- With the patch, license is output as:
  ...
  LICENSE = "United States Government Purpose & SwRI Proprietary"
  LIC_FILES_CHKSUM = "file://package.xml;beginline=11;endline=11;md5=8e473b26ead33548c945bdaca143a08d"
    ...

### DIFF
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -20,7 +20,6 @@ import string
 import sys
 import time
 
-from superflore.exceptions import UnknownLicense
 from superflore.exceptions import UnknownPlatform
 from superflore.rosdep_support import get_cached_index, resolve_rosdep_key
 from termcolor import colored
@@ -199,8 +198,8 @@ def get_license(l):
     elif re.search(pub_dom_re, l, f):
         return 'public_domain'
     else:
-        err('Could not match license "{0}".'.format(l))
-        raise UnknownLicense('bad license')
+        warn('Could not match license "{0}". Passing it through...'.format(l))
+        return l
 
 
 def resolve_dep(pkg, os, distro=None):
--- a/superflore/exceptions.py
+++ b/superflore/exceptions.py
@@ -13,11 +13,6 @@
 # limitations under the License.
 
 
-class UnknownLicense(Exception):
-    def __init__(self, message):
-        self.message = message
-
-
 class UnresolvedDependency(Exception):
     def __init__(self, message):
         self.message = message
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -14,7 +14,6 @@
 
 from rosinstall_generator.distro import get_package_names
 from superflore.exceptions import UnknownBuildType
-from superflore.exceptions import UnknownLicense
 from superflore.utils import err
 from superflore.utils import get_pkg_version
 from superflore.utils import info
@@ -75,10 +74,6 @@ def generate_installers(
             else:
                 changes.append('*{0} {1}*'.format(pkg, version))
             installers.append(pkg)
-        except UnknownLicense as ul:
-            err("{0}%: Unknown License '{1}'.".format(percent, str(ul)))
-            bad_installers.append(pkg)
-            failed = failed + 1
         except UnknownBuildType as ub:
             err(
                 "{0}%: Unknown Build type '{1}' for package '{2}'".format(
--- a/tests/test_generate_installers.py
+++ b/tests/test_generate_installers.py
@@ -16,7 +16,6 @@ import re
 
 from rosinstall_generator.distro import get_distro
 from superflore.exceptions import UnknownBuildType
-from superflore.exceptions import UnknownLicense
 from superflore.generate_installers import generate_installers
 import unittest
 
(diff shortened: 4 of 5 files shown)


andre-rosa:
@allenh1  Ptal

superflore/utils.py
```diff
@@ -198,6 +200,10 @@ def get_license(l):
         return 'Boost-1.0'
     elif re.search(pub_dom_re, l, f):
         return 'public_domain'
+    elif re.search(government_re, l, f):
+        return 'GPR'
+    elif re.search(proprietary_re, l, f):
+        return 'Proprietary'
```
allenh1: So I don't think we can pull this change in. Gentoo doesn't play nice with these strings. While we can certainly add the "GPR" license, the "Proprietary" string is rather meaningless.

At one point in time, I had an idea to package the license for any "Proprietary" package by fetching it from the repo, but this was rather low on the TODO-list for superflore. 

I don't know what Open Embedded has for this situation, but, from time to time, we have to add licenses to the overlay to support installation of certain packages. They go to [this folder](https://github.com/ros/ros-overlay/tree/master/licenses).

Maybe I'll add that feature to superflore soon (it would capture the LICENSE file from the repo, then it would add the new license to the overlay in the PR for the update).

allenh1: Upon looking at this package further, I found that it lacks a LICENSE file completely. I would file an issue on that repo and state that the repo lacks a LICENSE file, and should have one (so that we can clearly see the terms that are agreed upon by accepting the GPR license).

superflore/utils.py
```diff
@@ -155,6 +155,8 @@ def get_license(l):
     boost_re = '^(Boost)((.)*([1]))?'
     pub_dom_re = '^(Public(.)?Domain)'
     mit_re = '^MIT'
+    government_re = '.*Government.*'
```
tfoote: This is a little broad for capturing GPR in general. I also don't really consider GPR a license. As it is not something that applies to anyone else besides the US Government.  And it's specific to the deliverable provided by a contractor. It doesn't have carry on effects for future licensees or changes.

superflore/utils.py
tfoote: I think that doing something where if it doesn't match any of our pre-categorized licenses it would be classified as "Custom" or something catchall. And capturing the raw license file from the repo would be valuable.

There's a level that we're just trying to group the licenses for easier review. If we can't classify it we should just pass it through unclassified.

andre-rosa: Ptal at the modified patch that passes the licenses through unclassified when we can't recognize and categorize. I believe we'll not have a use for UnknownLicense exception if we'll pass it through when unrecognized, so please let me know if I should remove it.

Submitted https://github.com/swri-robotics/qt_metapackages/issues/1 asking for adding licensing files there.

tfoote:
This looks good to me. @allenh1 Does Gentoo require that they be specifically classified or from a certain list? If so we could always add an optional argument to this that would trigger the exception if the license can't be classified. And this could be used by Gentoo packaging, but not set by the OpenEmbedded invocation.

allenh1:
> Does Gentoo require that they be specifically classified or from a certain list?

Gentoo doesn't, but a lot of users do. In the past, we had this sort of rule in place (though we just copied the license string in that case). I think that, for now, having an "unclassified" license will be just fine.

I think we do a release after this & the rest of the preliminary open embedded PR's come in, then I'll get the automatic license pulling on the docker for the next superflore release.

superflore/generators/bitbake/yocto_recipe.py
```diff
@@ -146,9 +153,9 @@ def get_recipe_text(self, distributor, license_text):
             ret += 'LICENSE = "'
             ret += ' & '.join([get_license(l) for l in self.license]) + '"\n'
         ret += 'LIC_FILES_CHKSUM = "file://package.xml;beginline='
-        ret += str(self.license_line)
+        ret += str(self.license_line_begin)
         ret += ';endline='
-        ret += str(self.license_line)
+        ret += str(self.license_line_end)
```
allenh1: I think we are solving a different problem here -- please don't include such a fix in this PR.

superflore/utils.py
```diff
@@ -199,8 +199,8 @@ def get_license(l):
     elif re.search(pub_dom_re, l, f):
         return 'public_domain'
     else:
-        err('Could not match license "{0}".'.format(l))
-        raise UnknownLicense('bad license')
+        warn('Could not match license "{0}". Passing it through...'.format(l))
```
allenh1: This change requires some test changes. Please modify the appropriate tests.

superflore/generators/bitbake/yocto_recipe.py
andre-rosa: Done

superflore/utils.py
andre-rosa: Done

allenh1:
@andre-rosa lgtm. Thanks for iterating on this.

andre-rosa:
Thank you, @allenh1 !

PR #187, opened 2019-05-08
Fix gen_pkg_func failure not being recorded by generate_installers

### LINKED ISSUE #186: generate_installers() may record a failure in gen_pkg_func() as success
generate_installers() doesn't consider a few gen_pkg_func() failure situations like:
  - `InvalidPackage`: The manifest contains invalid XML: syntax error: line 1, column 0
    !!!! 61.7%: Failed to generate installer for package 'qb_chain'!
  - `NoPkgXml`: Could not fetch pkg!

Whenever gen_pkg_func() doesn't return an installer, it must be an error except when we want to preserve_existing , therefore make it record a failure when no installer is returned and we don't want to preserve_existing .

andre-rosa:
generate_installers() didn't consider a few gen_pkg_func() failure
situations like:

  - InvalidPackage:
    The manifest contains invalid XML: syntax error: line 1, column 0
    !!!! 61.7%: Failed to generate installer for package 'qb_chain'!
  - NoPkgXml: Could not fetch pkg!

Those are signaled by gen_pkg_func by returning no installer (None)
and no missing dependencies ([]).

Whenever gen_pkg_func() returns no installer, it must be an error except
when we want to preserve_existing.

Fixes #186 .

### DIFF
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -50,17 +50,18 @@ def generate_installers(
             current, current_info = gen_pkg_func(
                 overlay, pkg, distro, preserve_existing, *args
             )
-            if not current and current_info:
-                # we are missing dependencies
+            if not current:
+                if preserve_existing:
+                    # don't replace the installer
+                    succeeded = succeeded + 1
+                    continue
+                if current_info:
+                    # we are missing dependencies
+                    borkd_pkgs[pkg] = current_info
                 failed_msg = "{0}%: Failed to generate".format(percent)
                 failed_msg += " installer for package '%s'!" % pkg
                 err(failed_msg)
-                borkd_pkgs[pkg] = current_info
                 failed = failed + 1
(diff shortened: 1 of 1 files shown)


allenh1:
@andre-rosa I don't think I'm quite understanding what this patch does. Can you explain what your concerns are for running without this/what this blocks?

andre-rosa:
> @andre-rosa I don't think I'm quite understanding what this patch does. Can you explain what your concerns are for running without this/what this blocks?

@allenh1  When get_pkg_func() raises NoPkgXml it returns `None, []` (no installer, no missing dependency) and that gets recorded as a success generating the installer by generate_installer().

Patch handles the situation `if not current and not preserve_existing: record failure`. In that case, it'll not output `ok('Successfully generated an installer for package ...')` but instead `err('Failed to generate an installer for package ...')`.

PR #189, opened 2019-05-10
Retry XML downloads

### LINKED ISSUE #188: Sometimes a transient XML download failure aborts generation
Every now and then a transient failure when downloading XML files would abort the generation.

andre-rosa:
- Adds a retry mechanism to the XML downloads. It makes generation more
    robust upon transient failures of the download which occurs more often
    than not. Normally downloads work fine just by retrying them.
  - Fixes #188 .

### DIFF
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -295,11 +295,13 @@ def retry_on_exception(callback, *args, max_retries=5, num_retry=0,
     except Exception as e:
         if num_retry >= max_retries or max_retries < 0 or num_retry < 0:
             if error_msg:
-                err(f'{e} {error_msg} {num_retry}/{max_retries}')
+                err('{0} {1} {2}/{3}'.format(str(e), error_msg,
+                    num_retry, max_retries))
             raise e from None
         if num_retry > 0:
             if retry_msg:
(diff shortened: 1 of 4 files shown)


andre-rosa:
@allenh1 << ptal

superflore/utils.py
```diff
@@ -286,3 +286,18 @@ def url_to_repo_org(url):
...
+            ret = callback(*args)
+        except Exception as e:
+            if retry == retries:
+                if error_msg:
+                    err(f'{error_msg} {e} {retry}/{retries}')
+                raise e
+            if retry_msg:
+                warn(f'{retry_msg} {e} {retry}/{retries}...')
+        else:
+            return ret
```
tfoote: It would be good to at least warn possibly error if the retry count is out of a valid range, (negative or zero) as the callback will never be tried.

It would probably make more sense to actually have it hit the callback first and then count retries. So an argument "retries=0" will still imply it tries once and then quits, which is 0 retries.

andre-rosa: Thanks @tfoote << Ptal

andre-rosa:
@tfoote << ptal

Issue #193, opened 2019-06-04
Recent Gentoo generations aren't listing missing dependencies


andre-rosa:
Noticed originally at https://github.com/ros/ros-overlay/pull/805 by @allenh1 .

It may have been caused by the fix in https://github.com/ros-infrastructure/superflore/commit/7d1941508999f1fc560f2d2e3e7e86f83d2f3bd7 that changed slightly the behavior Gentoo was relying upon.

andre-rosa:
Fixed by #194 .

PR #194, opened 2019-06-04
Check missing dependencies before checking preserve_existing

andre-rosa:
- Gentoo expects first checking for missing dependencies before
    considering preserve_existing, so keep checks in that order.


andre-rosa:
@allenh1  << Could you please try this and check if it fixes #193 ?

allenh1:
@andre-rosa that definitely seems to have done the trick. Thanks!

andre-rosa:
Closes #193

PR #199, opened 2019-06-13
Fix CI

allenh1:
Update GPG key for ROS buildfarm

Signed-off-by: Hunter L. Allen <hunterlallen@protonmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ python:
   - "pypy3"
 before_install:
   - sudo echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" | sudo tee -a /etc/apt/sources.list.d/ros-latest.list
-  - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
-  - sudo apt-get update -qq
-  - sudo apt-get install dpkg -y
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,10 +9,10 @@ python:
   - "pypy3"
 before_install:
   - sudo echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" | sudo tee -a /etc/apt/sources.list.d/ros-latest.list
-  - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
-  - sudo apt-get update -qq
-  - sudo apt-get install dpkg -y
-  - sudo apt-get install -y python-rosdep
+  - sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
+  - sudo apt update -qq
```
tfoote: Apt is more user friendly but they're not promising a stable api, in general for scripting I still recommend using apt-get.

allenh1: @tfoote That makes sense -- I'm still not sure what the difference is. I've just read stuff saying 'cool kids use apt now' and thought we might as well.

I'll revert that portion in the release PR, I think, since I'll rebase on top of this one anyway.

Issue #202, opened 2019-06-15
Update OpenEmbedded support up to spec r35


andre-rosa:
- Add support for changes in r34..r35

andre-rosa:
Superseded by #204.

herb-kuta-lge:
@andre-rosa < This issue can be closed now.

PR #203, opened 2019-06-15
Generate superflore-datetime.inc

andre-rosa:
- As per spec r35


andre-rosa:
Fixes #202

andre-rosa:
@allenh1  << Ptal

andre-rosa:
Superseded by #205 .

PR #208, opened 2019-06-28
Retry pushing when wrong credentials are provided or there's an error

### LINKED ISSUE #207: Superflore aborts if wrong credentials are passed when pushing
After doing a Superflore generation, if wrong credentials are passed when doing the github push, it'll abort the whole process.

Instead, we should retry if there was an error when pushing, avoiding losing all the generated artefacts.

andre-rosa:
Fixes #207 .


andre-rosa:
@allenh1 << Ptal

Issue #225, opened 2019-09-03
superflore-gen-oe-recipes: Show recipe names instead of ROS package names in commit and PR messsages


herb-kuta-lge:
What's being committed are recipes, so it's better to show their names instead of the ROS package names in the commit and PR messages.

shr-project:
Fixed by https://github.com/ros-infrastructure/superflore/pull/235

PR #235, opened 2019-11-15
Show OpenEmbedded recipe names in commit and PR messages (#225)

herb-kuta-lge:
What's being committed are recipes, so show their names instead of the
ROS package names in the commit and PR messages.

### DIFF
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -47,7 +47,7 @@ def generate_installers(
         version = get_pkg_version(distro, pkg, **kwargs)
         percent = '%.1f' % (100 * (float(i) / total))
         try:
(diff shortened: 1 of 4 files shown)


allenh1:
@herb-kuta-lge seems to be some test failures -- I'll take a closer look after those get resolved.

herb-kuta-lge:
>  I'll take a closer look after those get resolved.

@allenh1 Fixed.

shr-project:
Hi @allenh1 is there anything blocking this PR? We're almost ready to publish updated meta-ros with recipes generated with superflore modified with all these open PRs and it would be nice to have it in superflore/master before we do.

Issue #230, opened 2019-11-15
superflore-gen-oe-recipes: -native packages are missing from the list of platform package dependencies


herb-kuta-lge:
`ROS_SUPERFLORE_GENERATED_PLATFORM_PACKAGE_DEPENDENCIES` is a list of package names not recipe names => it should include `-native` packages.

herb-kuta-lge:
Resolved by PR #237 .

PR #237, opened 2019-11-15
Include -native pkgs in list of OpenEmbedded platform dependencies (#230)

herb-kuta-lge:
ROS_SUPERFLORE_GENERATED_PLATFORM_PACKAGE_DEPENDENCIES is a list of
package names, not a list of recipes => add -native dependencies to it.


herb-kuta-lge:
An OE build target with a `-native` suffix means it's a "build tool", ie, it's to be built for the native instruction set of the build machine and will be executed on it during the build. A recipe can be "extended" to build a native version by appending `native` to `BBCLASSEXTEND`.

allenh1:
@herb-kuta-lge Alright, good to know. I think I'll just merge this then, since it's passing CI.

andre-rosa:
Looks good to me also, thank you both!

Issue #239, opened 2019-11-15
CI uses EOL Ubuntu release 'trusty'


herb-kuta-lge:
Ubuntu **trusty** has reached its EOL => `.travis.yml` should be updated to use a newer release, eg, **bionic**.

allenh1:
👍 agreed

allenh1:
Fixed in #241

PR #241, opened 2019-11-15
Upgrade CI

allenh1:
Upgrade ubuntu to bionic, and upgrade python versions to 3.7 and 3.8

Signed-off-by: Hunter L. Allen <hunterlallen@protonmail.com>

---

Connects to #239

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ python:
   - "3.6"
   - "3.7"
   # PyPy versions
-  - "pypy3"
+  # TODO(allenh1): Re-enable pypy3 when available for ubuntu 18.04
+  # - "pypy3"
 before_install:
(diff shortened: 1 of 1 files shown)


allenh1:
@tfoote do you know what's wrong with `rosdep` and python 3.8? It seems to be unable to detect the os?

tfoote:
I suspect something changed in the rospkg os_detect module that isn't compatible anymore.

allenh1:
> I suspect something changed in the rospkg os_detect module that isn't compatible anymore.

Odd. I'll keep it on 3.6 and 3.7 then, I think.

allenh1:
@tfoote I had to remove `pypy3`, unfortunately. Do you think it's too soon to jump to bionic? I feel like it definitely shouldn't be, but I could go to an earlier release if desired (maybe xenial).

Issue #226, opened 2019-09-03
Use "recipe"/"ebuild" instead of "installer" in log messages


herb-kuta-lge:
The term "installer" doesn't make any sense in the context of OpenEmbedded; what superflore-gen-oe-recipes generates are recipes.

allenh1:
@herb-kuta-lge I think I chose "installer" because it seemed the most neutral. At this point, however, it feels more misleading.

Gentoo should be changed to "ebuild" as well.

shr-project:
Fixed by https://github.com/ros-infrastructure/superflore/pull/243

PR #243, opened 2019-11-15
 Use "ebuild"/"recipe" instead of "installer" (#226)

herb-kuta-lge:
Use the terms "ebuild" and "recipe" instead of "installer" except in
names that are meant to refer to both.

### DIFF
--- a/superflore/generate_installers.py
+++ b/superflore/generate_installers.py
@@ -38,8 +38,9 @@ def generate_installers(
     bad_installers = []
     succeeded = 0
     failed = 0
+    what_generating = 'recipe' if kwargs.get('is_oe', False) else 'ebuild'
 
-    info("Generating installers for distro '%s'" % distro_name)
+    info("Generating %ss for distro '%s'" % (what_generating, distro_name))
     for i, pkg in enumerate(sorted(pkg_names[0])):
         if 'skip_keys' in kwargs and pkg in kwargs['skip_keys']:
             warn("Package '%s' is in skip-keys list, skipping..." % pkg)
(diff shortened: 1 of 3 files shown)


superflore/generate_installers.py
```diff
@@ -38,16 +38,17 @@ def generate_installers(
     bad_installers = []
     succeeded = 0
     failed = 0
+    what_generating = 'recipe' if kwargs['is_oe'] else 'ebuild'
```
shr-project: This should be:kwargs.get('is_oe', False)

superflore/generate_installers.py
```diff
@@ -59,19 +60,19 @@ def generate_installers(
                     succeeded += 1
                     continue
                 failed_msg = "{0}%: Failed to generate".format(percent)
-                failed_msg += " installer for package '%s'!" % pkg
+                failed_msg += " %s for package '%s'!" % (what_generating, pkg)
                 err(failed_msg)
                 failed = failed + 1
                 continue
-            success_msg = 'Successfully generated installer for package'
+            success_msg = 'Successfully generated %s for package' % what_generating
```
shr-project: this line is too long

superflore/generators/ebuild/gen_packages.py
```diff
@@ -75,12 +75,12 @@ def regenerate_pkg(overlay, pkg, distro, preserve_existing=False):
         )
         overlay.repo.remove_file(manifest_file)
     try:
-        current = gentoo_installer(distro, pkg, has_patches)
+        current = ebuild(distro, pkg, has_patches)
```
shr-project: This should be gentoo_ebuild(

shr-project:
+1 all my comments were resolved

Issue #250, opened 2019-11-19
Create new commit when bitbake installer changes some other files without changing any generated recipes


shr-project:
It generates couple other files which aren't covered by num_changes variable (like generated distro, mappings).

Re-generating the recipes from the same index file but e.g. with updated rosdistro or some superflore modification can cause changes only to these other files without changing anything in the recipes themselves. But we still want to commit these changes with regular superflore commit.

PR #251, opened 2019-11-19
bitbake: create an commit when there are some other changes in generated files (#250)

shr-project:
* even when there aren't any changes in generated recipes

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -209,9 +209,14 @@ def main():
 
         if num_changes == 0:
             info('ROS distro is up to date.')
-            info('Exiting...')
-            clean_up()
-            sys.exit(0)
+            summary = overlay.get_change_summary()
+            if len(summary) == 0:
+                info('Exiting...')
+                clean_up()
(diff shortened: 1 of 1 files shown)


shr-project:
Fixes #250

superflore/generators/bitbake/run.py
```diff
@@ -209,9 +209,13 @@ def main():
...
-            info('Exiting...')
-            clean_up()
-            sys.exit(0)
+            summary = overlay.get_change_summary()
+            if len(summary) == 0:
+                info('Exiting...')
+                clean_up()
+                sys.exit(0)
+            else:
+                info('But there are some changes in other regenerated files')
```
allenh1: I would really prefer that this refer to the type of files that did change, and why they changed. Doesn't need to be super detailed or anything, just something along the lines of "Commiting changes to [what] because [why]".

shr-project:
Hi @allenh1 is there anything blocking this PR? We're almost ready to publish updated meta-ros with recipes generated with superflore modified with all these open PRs and it would be nice to have it in superflore/master before we do.

Issue #252, opened 2019-11-19
Don't write cache.yaml and cache.diffme files from superflore


shr-project:
Updating these 2 files was moved to new script in meta-ros ros-generate-cache.sh and superflore just generates recipes from existing cache.yaml.

PR #254, opened 2019-11-19
yocto_recipe.py: don't create cache.yaml and cache.diffme (#252)

shr-project:
* ros-generate-cache.sh already does that and superflore then just
  generates recipes from cache.yaml file updated by
  ros-generate-cache.sh
* fixes: https://github.com/ros-infrastructure/superflore/issues/252
* needs: https://github.com/ros/meta-ros/pull/643

### DIFF
--- a/superflore/generators/bitbake/ros_meta.py
+++ b/superflore/generators/bitbake/ros_meta.py
@@ -41,7 +41,6 @@ class RosMeta(object):
         info('Adding changes...')
         self.repo.git.add('generated-recipes-{0}'.format(distro))
         self.repo.git.add('conf/ros-distro/include/{0}/*.inc'.format(distro))
-        self.repo.git.add('files/{0}/cache.*'.format(distro))
         self.repo.git.add('files/{0}/rosdep-resolve.yaml'.format(distro))
         self.repo.git.add(
             'files/{0}/newer-platform-components.list'.format(distro))
@@ -68,7 +67,6 @@ class RosMeta(object):
         return '\n'.join([
             sep, self.repo.git.status('--porcelain'), sep,
             self.repo.git.diff('conf'), sep, self.repo.git.diff(
-                'files/*/cache.diffme',
                 'files/*/newer-platform-components.list',
                 'files/*/rosdep-resolve.yaml'
             ),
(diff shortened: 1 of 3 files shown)


shr-project:
> The `cache.yaml` that **ros-generate-cache** generated is only used by **superflore** if`ROSDISTRO_INDEX_URL` is in its environment => if it isn't, it should continue to generate `cache.diffme`.

So you want to keep this block and only call yoctoRecipe.generate_distro_cache(_repo, args.ros_distro) conditionally on ROSDISTRO_INDEX_URL being set? I don't mind changing that, but then the sed replacements will be in 2 places (ros-generate-cache.sh and superflore/generators/bitbake/yocto_recipe.py) and with ros-generate-recipes.sh always setting ROSDISTRO_INDEX_URL I wasn't expecting anyone to not use that.

herb-kuta-lge:
>  sed replacements will be in 2 places 

What I've done in the past to handle this is add a comment in each place that tells the reader making a change the other place that also needs to be updated.

>  with ros-generate-recipes.sh always setting ROSDISTRO_INDEX_URL I wasn't expecting anyone to not use that.

Then **superflore-gen-oe-recipes** should fail if `ROSDISTRO_INDEX_URL`  isn't set to make sure that no one uses it by itself. It seems like it would be more flexible not require that, though.

shr-project:
> > sed replacements will be in 2 places
> 
> What I've done in the past to handle this is add a comment in each place that tells the reader making a change the other place that also needs to be updated.
> 
> > with ros-generate-recipes.sh always setting ROSDISTRO_INDEX_URL I wasn't expecting anyone to not use that.
> 
> Then **superflore-gen-oe-recipes** should fail if `ROSDISTRO_INDEX_URL` isn't set to make sure that no one uses it by itself. It seems like it would be more flexible not require that, though.

If you don't want to use cache.yaml and rosdistro files stored in meta-ros (e.g. completely ignore ros-generate-cache.sh part), then you can still call superflore directly and it works (I've just tried to regen eloquent from latest index). We shouldn't accept such change to meta-ros, because it didn't correctly update the cache.yaml nor cache.diffme (and was created with unknown rosdistro mapping), but it still might be useful for someone locally (e.g. to see what's different in latest index without updating cache and rosdistro in meta-ros), why should we make ROSDISTRO_INDEX_URL mandatory like ros-generate-recipes.sh does?

herb-kuta-lge:
> it still might be useful for someone locally (e.g. to see what's different in latest index without updating cache and rosdistro in meta-ros

OK. I withdraw my request.

herb-kuta-lge:
@allenh1 < Ready to merge.

PR #262, opened 2019-12-17
Use "git describe" to append a local version label

shr-project:
and small fix for superflore-datetime.inc.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,49 @@
 import sys
 from setuptools import find_packages, setup
 
+"""
+Returns <public_version> appended with a PEP-440 compliant local version label
+(see https://www.python.org/dev/peps/pep-0440/#local-version-identifiers). The
+local version label is based on the output from
+"git describe --tags --dirty --broken". Nothing is appended if:
+- Git is not available, or
+- "git describe" might not be operating on a branch of superflore.git, or
+- there is no tag for <public_version>, or
+- the HEAD of the current branch is coincident with the tag for <public_version>.
+
+NB. Not using https://pypi.org/project/setuptools-git-version/ because it passes
+"--long" to "git describe" and doesn't pass "--broken".
+"""
+def append_local_version_label(public_version):
+    try:
+        from git import Repo
+        from os import getcwd
+        from os.path import join, samefile
+
+        repo = Repo()
+        """
+        If we're been copied under the working dir of some other Git repo,
+        "git describe" won't return what we're expecting, so don't append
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,6 +2,23 @@
 
 import sys
 from setuptools import find_packages, setup
+import git
+from git import Repo
```
allenh1: I think you can strike the `import git` line, since `from git import Repo` should cover it.

setup.py
```diff
@@ -24,7 +41,7 @@
 
 setup(
     name='superflore',
-    version='0.3.0',
+    version=get_version_from_git_describe(),
```
allenh1: @tfoote does this disrupt anything for releases?

allenh1:
@tfoote please review this one -- you know quite a bit more than I do about this.

shr-project:
@herb-kuta-lge FYI

setup.py
herb-kuta-lge: The step that previously created a commit to change the value of `version` is replaced by one that adds a tag. @allenh1 < Would this be too much of a disruption?

shr-project: I've dropped this controversial commit from the PR and added various other improvements instead. Please re-review.

herb-kuta-lge: Since I intend to someday propose an alternative implementation that addresses Tully's objections, I now think it would make more sense to create a new PR for the bitbake generator changes and just have the setup.py commit be in this PR.

shr-project:
@herb-kuta-lge requested to split this pull request into smaller ones with separate issue tickets and to leave this original one for setup.py change he plans to update with different implementation based on @tfoote's feedback. I've created:
https://github.com/ros-infrastructure/superflore/pull/265
https://github.com/ros-infrastructure/superflore/pull/266
https://github.com/ros-infrastructure/superflore/pull/267
https://github.com/ros-infrastructure/superflore/pull/268
https://github.com/ros-infrastructure/superflore/pull/269
https://github.com/ros-infrastructure/superflore/pull/270
to cover the other changes which were here yesterday.

herb-kuta-lge:
@tfoote < Please review the latest implementation. I believe it addresses all of your concerns.

setup.py
```diff
@@ -2,6 +2,51 @@
 
 import sys
 from setuptools import find_packages, setup
+import git
```
allenh1: is this line needed?

herb-kuta-lge: No; I'll remove it.

PR #265, opened 2020-02-05
superflore-gen-oe-recipes: Delete all generated files when regenerating all recipes

shr-project:
* the overlay.repo.remove_file(existing, True) call in bitbake/gen_packages.py
  works fine, but only when the new recipe is being generated in the same directory
  as the previous version:
  existing = glob.glob('{}_*.bb'.format(prefix))
  where prefix is:
  prefix = '{0}/meta-ros{1}-{2}/generated-recipes/{3}/{4}'
  which will result in multiple recipes for the same package when the
  package is moved to different prefix like pcl-conversions did for
  dashing and eloquent:
```
  meta-ros$ find . -name pcl-conversions\*bb
  ./meta-ros2-eloquent/generated-recipes/perception-pcl/pcl-conversions_2.1.0-1.bb
  ./meta-ros2-eloquent/generated-recipes/pcl-conversions/pcl-conversions_2.0.0-1.bb
  ./meta-ros1-melodic/generated-recipes/perception-pcl/pcl-conversions_1.7.0-2.bb
  ./meta-ros2-crystal/generated-recipes/pcl-conversions/pcl-conversions_2.0.0.bb
  ./meta-ros2-dashing/generated-recipes/perception-pcl/pcl-conversions_2.0.0-1.bb
  ./meta-ros2-dashing/generated-recipes/pcl-conversions/pcl-conversions_2.0.0-1.bb
```
  and especially for dashing where the version is identical it's
  undeterministic which recipe bitbake will use. And the recipe wasn't
  just moved to different directory, it's quite different:
```
diff ./meta-ros2-dashing/generated-recipes/perception-pcl/pcl-conversions_2.0.0-1.bb ./meta-ros2-dashing/generated-recipes/pcl-conversions/pcl-conversions_2.0.0-1.bb
9c9
< AUTHOR = "Paul Bovbel <paul@bovbel.com>"
---
> AUTHOR = "Chris Lalancette <clalancette@openrobotics.org>"
14c14
< LIC_FILES_CHKSUM = "file://package.xml;beginline=16;endline=16;md5=d566ef916e9dedc494f5f793a6690ba5"
---
> LIC_FILES_CHKSUM = "file://package.xml;beginline=12;endline=12;md5=d566ef916e9dedc494f5f793a6690ba5"
16c16
< ROS_CN = "perception_pcl"
---
> ROS_CN = "pcl_conversions"
19a20
>     builtin-interfaces \
21d21
<     message-filters \
23,24d22
<     pcl-msgs \
<     rclcpp \
33a32
>     builtin-interfaces \
35d33
<     message-filters \
37,38d34
<     pcl-msgs \
<     rclcpp \
46,47d41
<     libeigen \
<     message-filters \
49,52d42
<     pcl-msgs \
<     rclcpp \
<     sensor-msgs \
<     std-msgs \
67c57
< # matches with: https://github.com/ros2-gbp/perception_pcl-release/archive/release/dashing/pcl_conversions/2.0.0-1.tar.gz
---
> # matches with: https://github.com/ros2-gbp/pcl_conversions-release/archive/release/dashing/pcl_conversions/2.0.0-1.tar.gz
69,70c59,60
< SRC_URI = "git://github.com/ros2-gbp/perception_pcl-release;${ROS_BRANCH};protocol=https"
< SRCREV = "0f550985da8bd73fd7451350c63e1ce96e78c955"
---
> SRC_URI = "git://github.com/ros2-gbp/pcl_conversions-release;${ROS_BRANCH};protocol=https"
> SRCREV = "8d0d0ef7d5d253f11fbf52388d64fad59e46377a"
```
* grep in the current build logs shows that the "older"
  meta-ros2-dashing/generated-recipes/pcl-conversions/pcl-conversions_2.0.0-1.bb
  was currently being used in dashing instead of "newer"
  meta-ros2-dashing/generated-recipes/perception-pcl/pcl-conversions_2.0.0-1.bb

* corresponding rosdistro change for dashing:
  https://github.com/ros/rosdistro/commit/19b06c44df69f9bf44dd7ab1b74236e9e386c848
  and eloquent:
  https://github.com/ros/rosdistro/commit/67c3ef53979ea8d85254bdfc427b142eda69fb57

### DIFF
--- a/superflore/generators/bitbake/gen_packages.py
+++ b/superflore/generators/bitbake/gen_packages.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import glob
 import os
 
 from catkin_pkg.package import InvalidPackage
@@ -28,6 +27,7 @@ from superflore.utils import get_pkg_version
 from superflore.utils import make_dir
 from superflore.utils import ok
 from superflore.utils import retry_on_exception
+from superflore.utils import warn
 
 org = "Open Source Robotics Foundation"
 
@@ -49,22 +49,58 @@ def regenerate_pkg(
     component_name = yoctoRecipe.convert_to_oe_name(
         distro.release_packages[pkg].repository_name)
     recipe = yoctoRecipe.convert_to_oe_name(pkg)
-    # check for an existing recipe
-    prefix = '{0}/meta-ros{1}-{2}/generated-recipes/{3}/{4}'.format(
-        repo_dir,
+    # check for an existing recipe which was removed by clean_ros_recipe_dirs
+    prefix = 'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb'.format(
         yoctoRecipe._get_ros_version(distro.name),
         distro.name,
-        component_name,
-        recipe,
+        recipe
     )
-    existing = glob.glob('{}_*.bb'.format(prefix))
+    existing = overlay.repo.git.status('--porcelain', '--', prefix)
+    if existing:
+        # The git status --porcelain output will look like this:
+        # D  meta-ros2-eloquent/generated-recipes/variants/ros-base_0.8.3-1.bb
+        # we want just the path with filename
+        if len(existing.split('\n')) > 1:
+            warn('More than 1 recipe was output by "git status --porcelain '
+                 'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                 .format(
+                     yoctoRecipe._get_ros_version(distro.name),
+                     distro.name,
+                     recipe,
+                     existing))
+        if existing.split()[0] != 'D':
+            err('Unexpected output from "git status --porcelain '
+                'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                .format(
+                    yoctoRecipe._get_ros_version(distro.name),
+                    distro.name,
+                    recipe,
+                    existing))
+
+        existing = existing.split()[1]
+    else:
+        # If it isn't shown in git status, it could still exist as normal
+        # unchanged file when --only option is being used
+        import glob
+        existing = glob.glob('{0}/{1}'.format(repo_dir, prefix))
+        if existing:
+            if len(existing) > 1:
+                err('More than 1 recipe was output by "git status '
+                    '--porcelain '
+                    'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                    .format(
+                        yoctoRecipe._get_ros_version(distro.name),
+                        distro.name,
+                        recipe,
+                        existing))
+            existing = existing[0]
+
     previous_version = None
     if preserve_existing and existing:
         ok("recipe for package '%s' up to date, skipping..." % pkg)
         yoctoRecipe.not_generated_recipes.add(pkg)
         return None, [], None
     elif existing:
-        existing = existing[0]
         overlay.repo.remove_file(existing, True)
         idx_version = existing.rfind('_') + len('_')
         previous_version = existing[idx_version:].rstrip('.bb')
--- a/superflore/generators/bitbake/ros_meta.py
+++ b/superflore/generators/bitbake/ros_meta.py
@@ -29,16 +29,23 @@ class RosMeta(object):
             info('Creating new branch {0}...'.format(self.branch_name))
             self.repo.create_branch(self.branch_name)
 
-    def clean_ros_recipe_dirs(self, distro=None):
-        if distro:
-            info(
-                'Cleaning up meta-ros{}-{}/generated-recipes directory...'
-                .format(yoctoRecipe._get_ros_version(distro), distro))
-            self.repo.git.rm('-rf', 'meta-ros{}-{}/generated-recipes'.format(
-                yoctoRecipe._get_ros_version(distro), distro))
-        else:
-            info('Cleaning up generated-recipes directories...')
-            self.repo.git.rm('-rf', 'generated-recipes')
+    def clean_ros_recipe_dirs(self, distro):
+        # superflore-change-summary.txt is no longer being generated since:
+        # https://github.com/ros-infrastructure/superflore/pull/273
+        # but remove it here to make sure it gets deleted when new distro
+        # release is being generated
+        files = 'meta-ros{0}-{1}/generated-recipes '\
+                'meta-ros{0}-{1}/conf/ros-distro/include/{1}/generated '\
+                'meta-ros{0}-{1}/files/{1}/generated/'\
+                'newer-platform-components.list '\
+                'meta-ros{0}-{1}/files/{1}/generated/rosdep-resolve.yaml '\
+                'meta-ros{0}-{1}/files/{1}/generated/'\
+                'superflore-change-summary.txt '.format(
+                    yoctoRecipe._get_ros_version(distro), distro)
+        info(
+            'Cleaning up:\n{0}'
+            .format(files))
+        self.repo.git.rm('-rf', '--ignore-unmatch', files)
 
     def commit_changes(self, distro, commit_msg):
         info('Commit changes...')
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -158,6 +158,7 @@ def main():
                 ok('Successfully synchronized repositories!')
                 sys.exit(0)
 
+            overlay.clean_ros_recipe_dirs(args.ros_distro)
             for adistro in selected_targets:
                 yoctoRecipe.reset()
                 distro = get_distro(adistro)


superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +48,32 @@ def regenerate_pkg(
...
+    if existing:
+        # The git status --porcelain output will look like this:
+        # D  meta-ros2-eloquent/generated-recipes/variants/ros-base_0.8.3-1.bb
+        # we want just the path with filename
+        existing = existing.split()[1]
+    if not existing:
+        # If it isn't shown in git status, it could still exist as normal
+        # unchanged file when --only option is being used
+        import glob
+        existing = glob.glob('{}/{}'.format(repo_dir, prefix))
```
herb-kuta-lge: '{}/{}' -> '{0}/{1}'

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +48,32 @@ def regenerate_pkg(
...
         distro.name,
-        component_name,
         recipe,
     )
-    existing = glob.glob('{}_*.bb'.format(prefix))
+    existing = overlay.repo.git.status('--porcelain', '--', prefix)
+    if existing:
+        # The git status --porcelain output will look like this:
+        # D  meta-ros2-eloquent/generated-recipes/variants/ros-base_0.8.3-1.bb
+        # we want just the path with filename
```
herb-kuta-lge: It might be a good idea to confirm that the output is what you're expecting.

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +48,32 @@ def regenerate_pkg(
...
+        existing = existing.split()[1]
+    if not existing:
```
herb-kuta-lge: Can line 62 ever result it existing evaluting to False? If not, then shouldn't this line be `else:`?

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +48,32 @@ def regenerate_pkg(
...
+        # If it isn't shown in git status, it could still exist as normal
+        # unchanged file when --only option is being used
+        import glob
+        existing = glob.glob('{}/{}'.format(repo_dir, prefix))
+        if existing:
```
herb-kuta-lge: What should happen if the glob returns more than one?

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +48,32 @@ def regenerate_pkg(
...
+    existing = overlay.repo.git.status('--porcelain', '--', prefix)
+    if existing:
+        # The git status --porcelain output will look like this:
+        # D  meta-ros2-eloquent/generated-recipes/variants/ros-base_0.8.3-1.bb
+        # we want just the path with filename
+        existing = existing.split()[1]
+    if not existing:
+        # If it isn't shown in git status, it could still exist as normal
+        # unchanged file when --only option is being used
+        import glob
```
herb-kuta-lge: @allenh1 < Is this style correct, or should all `import`-s be all the top of the files?

superflore/generators/bitbake/ros_meta.py
```diff
@@ -29,30 +29,39 @@ def __init__(
...
-            info(
-                'Cleaning up meta-ros{}-{}/generated-recipes directory...'
-                .format(yoctoRecipe._get_ros_version(distro), distro))
-            self.repo.git.rm('-rf', 'meta-ros{}-{}/generated-recipes'.format(
-                yoctoRecipe._get_ros_version(distro), distro))
-        else:
-            info('Cleaning up generated-recipes directories...')
-            self.repo.git.rm('-rf', 'generated-recipes')
+    def clean_ros_recipe_dirs(self, distro):
+        info(
```
herb-kuta-lge: So that the list of generated files/directories to be removed is in a single place, generate a list of them and then use it in the `info()` and `git rm -rf` calls. (A chance to use `map()`!)

superflore/generators/bitbake/ros_meta.py
```diff
@@ -74,17 +83,28 @@ def get_file_revision_logs(self, *file_path):
     def get_change_summary(self, distro):
         self.repo.git.add('meta-ros{0}-{1}/generated-recipes'.format(
             yoctoRecipe._get_ros_version(distro), distro))
+        self.repo.git.add('meta-ros{0}-{1}/conf/ros-distro/include/{1}/'
```
herb-kuta-lge: Why was this moved here? It seems confusing that the function `commit_changes()` doesn't commit all of the changes.

superflore/generators/bitbake/ros_meta.py
```diff
@@ -74,17 +83,28 @@ def get_file_revision_logs(self, *file_path):
...
+        self.repo.git.add('meta-ros{0}-{1}/files/{1}/generated/'
+                          'newer-platform-components.list'.format(
+                              yoctoRecipe._get_ros_version(distro), distro))
         sep = '-' * 5
         return '\n'.join([
             sep,
             self.repo.git.status('--porcelain'),
             sep,
             self.repo.git.diff(
+                'HEAD',
```
herb-kuta-lge: This is needed because of the newly added `git add`-s, right?

superflore/generators/bitbake/gen_packages.py
shr-project: The same as before with pcl-conversions, but it's a bit unclear how --only should behave as it isn't implemented yet. Now it will remove only the first one found, but on the other hand, since this change there shouldn't be meta-ros with multiple recipes for the same component, so using --only on correctly created meta-ros shouldn't find 2nd recipe, I'll add an error when there are 2 and warning in the git status above (so that pcl-conversions trigger just an warning).

superflore/generators/bitbake/gen_packages.py
shr-project: > What should happen if the glob returns more than one?

The same as before with pcl-conversions, but it's a bit unclear how --only should behave as it isn't implemented yet. Now it will remove only the first one found, but on the other hand, since this change there shouldn't be meta-ros with multiple recipes for the same component, so using --only on correctly created meta-ros shouldn't find 2nd recipe, I'll add an error when there are 2 and warning in the git status above (so that pcl-conversions trigger just an warning).

superflore/generators/bitbake/ros_meta.py
shr-project: yes

superflore/generators/bitbake/ros_meta.py
shr-project: I've moved this to separate PR to avoid confusion.

shr-project:
> Drop this and what follows. I don't see how showing how different the recipes are helps further explain the need for the change.

I disagree, It shows why it's important to implement this change.

herb-kuta-lge:
> I disagree, It shows why it's important to implement this change.

shr-project:
> Wouldn't still be important to implement it even if the recipes were identical?

No it wouldn't be important, because bitbake doesn't care from which directory the recipe was used, but when the built output is very different, then it's important to use the new recipe.

herb-kuta-lge:
> even if the recipes were identical

Actually, if a recipe moves to a different subdirectory, it's `ROS_CN` setting will be different, so the recipes will never be identical.

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +49,58 @@ def regenerate_pkg(
...
-        recipe,
+        recipe
     )
-    existing = glob.glob('{}_*.bb'.format(prefix))
+    existing = overlay.repo.git.status('--porcelain', '--', prefix)
+    if existing:
+        # The git status --porcelain output will look like this:
+        # D  meta-ros2-eloquent/generated-recipes/variants/ros-base_0.8.3-1.bb
+        # we want just the path with filename
+        if len(existing.split()) > 2:
```
herb-kuta-lge: It would be clearer to write `if len(existing) > 1:`

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +49,58 @@ def regenerate_pkg(
...
+            warn('More than 1 recipe was found in ""git status --porcelain '
+                 'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                 .format(
+                     yoctoRecipe._get_ros_version(distro.name),
+                     distro.name,
+                     recipe,
+                     existing))
```
herb-kuta-lge: I think this should be considered an assertion failure and therefore, the program should abort.

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +49,58 @@ def regenerate_pkg(
...
+        if existing.split()[0] != 'D':
+            err('Unexpected output of "git status --porcelain '
+                'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                .format(
+                    yoctoRecipe._get_ros_version(distro.name),
+                    distro.name,
+                    recipe,
+                    existing))
```
herb-kuta-lge: I think this should be considered an assertion failure and therefore, the program should abort.

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +49,58 @@ def regenerate_pkg(
...
+        if existing:
+            if len(existing) > 1:
+                err('More than 1 recipe was found in ""git status '
+                    '--porcelain '
+                    'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                    .format(
+                        yoctoRecipe._get_ros_version(distro.name),
+                        distro.name,
+                        recipe,
+                        existing))
```
herb-kuta-lge: I think this should be considered an assertion failure and therefore, the program should abort.

superflore/generators/bitbake/ros_meta.py
```diff
@@ -29,16 +29,36 @@ def __init__(
...
-            info('Cleaning up generated-recipes directories...')
-            self.repo.git.rm('-rf', 'generated-recipes')
+    def clean_ros_recipe_dirs(self, distro):
+        info(
+            'Cleaning up:\n'
+            'meta-ros{0}-{1}/generated-recipes\n'
+            'meta-ros{0}-{1}/conf/ros-distro/include/{1}/generated\n'
+            'meta-ros{0}-{1}/files/{1}/generated/'
+            '{{newer-platform-components.list,rosdep-resolve.yaml,'
+            'superflore-change-summary.txt}}\n'
```
herb-kuta-lge: As of https://github.com/ros-infrastructure/superflore/pull/273, `superflore-change-summary.txt` is no longer generated.

superflore/generators/bitbake/ros_meta.py
```diff
@@ -29,16 +29,36 @@ def __init__(
...
+            '-rf',
+            'meta-ros{0}-{1}/files/{1}/generated/'
+            'newer-platform-components.list'
+            .format(yoctoRecipe._get_ros_version(distro), distro))
+        self.repo.git.rm(
+            '-rf',
+            'meta-ros{0}-{1}/files/{1}/generated/'
+            'rosdep-resolve.yaml'
+            .format(yoctoRecipe._get_ros_version(distro), distro))
+        self.repo.git.rm(
```
herb-kuta-lge: As of https://github.com/ros-infrastructure/superflore/pull/273, `superflore-change-summary.txt` is no longer generated.

Why not create a Python list of the files/directories to be removed and use `map()` apply `git rm -rf` to them?

superflore/generators/bitbake/gen_packages.py
shr-project: Isn't the string always longer than 1?

superflore/generators/bitbake/ros_meta.py
shr-project: But it still be removed so that we don't need to remove it manually when regenerating the recipes, right?

superflore/generators/bitbake/gen_packages.py
shr-project: But that will require to remove the already duplicated pcl recipe manually before running superflore, otherwise it would fail and abort.

superflore/generators/bitbake/gen_packages.py
herb-kuta-lge: Isn't `existing` the array of lines returned by `git status --porcelain` and aren't we expecting there to only be one line in it?

herb-kuta-lge: I've confirmed that what's returned by `git.status()` is a single string with embedded newlines => write as `if len(existing.split('\n')) > 1:`

superflore/generators/bitbake/gen_packages.py
herb-kuta-lge: Then please update the comment on lines 60-62 for the case when there is more than one line output. Is the first token guaranteed to always be `D`? If not, then line 71 will need correcting.

superflore/generators/bitbake/gen_packages.py
herb-kuta-lge: `output of` -> `output from`

superflore/generators/bitbake/gen_packages.py
herb-kuta-lge: `found in` -> `output by`

superflore/generators/bitbake/gen_packages.py
herb-kuta-lge: `found in` -> `output by`

superflore/generators/bitbake/ros_meta.py
herb-kuta-lge: Yes. Add a comment explaining this.

superflore/generators/bitbake/gen_packages.py
```diff
@@ -24,6 +23,7 @@
 from superflore.exceptions import NoPkgXml
 from superflore.generators.bitbake.yocto_recipe import yoctoRecipe
 from superflore.utils import err
+from superflore.utils import warn
 from superflore.utils import get_pkg_version
```
herb-kuta-lge: From Travis CI:
```
superflore/generators/bitbake/gen_packages.py:27:1: I100 Import statements are in the wrong order. 'from superflore.utils import get_pkg_version' should be before 'from superflore.utils import warn'
```

superflore/generators/bitbake/gen_packages.py
herb-kuta-lge: Since the linter-s haven't complained about it, I guess this is OK.

herb-kuta-lge:
>  I don't see how showing how different the recipes

> I disagree, It shows why it's important to implement this change.

The details of the differences are important, so perhaps show how different they are by piping the `diff` output into `diffstat`? (Use `diff -C 1` so that the filename appears in the `diffstat` output.)

superflore/generators/bitbake/gen_packages.py
shr-project: "More than 1 recipe was found in output by "git status..."
or
"More than 1 recipe was output by "git status..."
as -> suggests?

superflore/generators/bitbake/ros_meta.py
shr-project: Which PR is going to be merged first, because with this PR the superflore-change-summary.txt is still being generated.

superflore/generators/bitbake/ros_meta.py
shr-project: Which PR is going to be merged first, because with this PR the superflore-change-summary.txt is still being generated.

superflore/generators/bitbake/ros_meta.py
herb-kuta-lge: #273 will be merged first -- I've added to note to the PR description reminding us of this.

superflore/generators/bitbake/gen_packages.py
```diff
@@ -49,22 +49,58 @@ def regenerate_pkg(
...
+        if existing.split()[0] != 'D':
+            err('Unexpected output from "git status --porcelain '
+                'meta-ros{0}-{1}/generated-recipes/*/{2}_*.bb": "{3}"'
+                .format(
+                    yoctoRecipe._get_ros_version(distro.name),
+                    distro.name,
+                    recipe,
+                    existing))
+
+        existing = existing.split()[1]
```
herb-kuta-lge, line 80: If more than one recipe what output by `git status --porcelain`, is it always true that the first one is the one to process?

superflore/generators/bitbake/ros_meta.py
herb-kuta-lge: #273 has been merged.

shr-project:
> Comments explaining what would cause unexpected output from git status --porcelain and glob.

Can you please provide these comments?

I don't know what unexpected might have happen especially in combination with --only which isn't implemented yet AFAIK.

superflore/generators/bitbake/gen_packages.py
shr-project: Currently it's just used to show in commit message, I don't think it's important which one is used (as long as there was an warning about it and the commit message doesn't say that the component is new, we might sort them based on version and then choose the one with highest version - because that's probably the one bitbake was building before, but that seems unnecessary for commit message in cases like this which shouldn't ever happen again.

herb-kuta-lge:
> I don't know what unexpected might have happen especially in combination with --only which isn't implemented yet AFAIK.

herb-kuta-lge:
> I don't know what unexpected might have happen especially in combination with --only which isn't implemented yet AFAIK.

If it's unexpected, shouldn't it be considered an ASSERT that stops execution?

shr-project:
The output is verbose, but err() output is quite rare in it and contains even more important messages than this unexpected output, see the errors in last meta-ros regeneration:
http://paste.ubuntu.com/p/b6JbJyTp8d/

herb-kuta-lge:
OK, then let's leave it as is.

herb-kuta-lge:
@allenh1 < Now that this last open PR was been merged, could you release a v0.3.2 ?

allenh1:
@herb-kuta-lge That sounds like a good idea to me!

@tfoote FYI

PR #266, opened 2020-02-05
bitbake: Don't create a pull request

shr-project:
* there usually are some changes needed to manually adjust the .bbappends to build everything
  correctly with the updated recipes, prevent people from accidentally creating pull-requests
  on github before their changes and is ready to build (or even parse correctly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ should be followed to generate the OpenEmbedded recipes for `meta-ros`.
 $ superflore-gen-oe-recipes --help
 usage: Deploy ROS packages into OpenEmbedded Linux [-h]
                                                    [--ros-distro ROS_DISTRO]
-                                                   [--dry-run] [--pr-only] [--no-branch]
+                                                   --dry-run [--pr-only] [--no-branch]
                                                    [--output-repository-path OUTPUT_REPOSITORY_PATH]
                                                    [--only ONLY [ONLY ...]]
                                                    [--pr-comment PR_COMMENT]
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -189,8 +189,12 @@ This command will clone the `ros/meta-ros` repo into a subfolder under
 distro, commit them, and issue a pull request for `ros/meta-ros`. The
 `--ros-distro` flag must be supplied.
 
-If you don't want to issue a PR for `ros/meta-ros`, you should add
-the `--dry-run` flag. You can issue the PR later by using the `--pr-only` flag.
+Generating bitbake recipes without specifying `--dry-run` is not
```
herb-kuta-lge: Drop [ ] from --dry-run in usage on line 145.

Add an optional `require_dryrun=False` parameter to `get_parser()`, which passes its value to the `required` parameter to the `add_argument()` invocation for `--dry-run` (see https://docs.python.org/3.6/library/argparse.html#required).

superflore/generators/bitbake/run.py
```diff
@@ -58,6 +58,12 @@ def main():
     if not args.dry_run:
         if 'SUPERFLORE_GITHUB_TOKEN' not in os.environ:
             raise NoGitHubAuthToken()
+        err('Generating bitbake recipes without specifying --dry-run is not '
```
herb-kuta-lge: Move to be before the test for `SUPERFLORE_GITHUB_TOKEN`.

Actually, since we now require `--dry-run`, this should be change to an "assert" of faulty program logic.

shr-project: We can delete all this, because now people won't see this longer explanation about --dry-run, only:

Deploy ROS packages into OpenEmbedded Linux: error: the following arguments are required: --dry-run

herb-kuta-lge:
@allenh1 < Should an issue be created for this PR?

allenh1:
> @allenh1 < Should an issue be created for this PR?

No, I'm not too concerned about it. Feel free to squash & merge.

PR #272, opened 2020-03-23
bitbake/run.py: Require --ros-distro

shr-project:
* enforce the requirement in the parser like --dry-run does now

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -28,7 +28,6 @@ from superflore.utils import clean_up
 from superflore.utils import err
 from superflore.utils import file_pr
 from superflore.utils import gen_delta_msg
-from superflore.utils import get_distros_by_status
 from superflore.utils import get_pr_text
 from superflore.utils import get_utcnow_timestamp_str
 from superflore.utils import info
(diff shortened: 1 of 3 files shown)


superflore/generators/bitbake/run.py
```diff
@@ -71,14 +71,10 @@ def main():
...
-        warn('"{0}" distro detected...'.format(args.ros_distro))
-        selected_targets = [args.ros_distro]
-        preserve_existing = False
-    elif args.only:
-        parser.error('Invalid args! --only requires specifying --ros-distro')
-    if not selected_targets:
-        selected_targets = get_distros_by_status('active')
+    warn('"{0}" distro detected...'.format(args.ros_distro))
+    selected_targets = [args.ros_distro]
+    preserve_existing = False
```
herb-kuta-lge: `preserve_existing` should remain `True` if `args.only` is `True` => drop line 44 and replace 76 with: `preserve_existing = args.only`

superflore/generators/bitbake/run.py
```diff
@@ -71,14 +71,10 @@ def main():
...
-    elif args.ros_distro:
-        warn('"{0}" distro detected...'.format(args.ros_distro))
-        selected_targets = [args.ros_distro]
-        preserve_existing = False
-    elif args.only:
-        parser.error('Invalid args! --only requires specifying --ros-distro')
-    if not selected_targets:
-        selected_targets = get_distros_by_status('active')
+    warn('"{0}" distro detected...'.format(args.ros_distro))
+    selected_targets = [args.ros_distro]
```
herb-kuta-lge: Drop line 58 above.

Add comment above line 75:
```
"""
No longer supporting generation for multiple targets, but left the code in
place to handle them in case it might be needed again in the future.
"""
```

superflore/generators/bitbake/run.py
```diff
@@ -71,14 +71,10 @@ def main():
...
+    preserve_existing = False
+
```
herb-kuta-lge: Drop extraneous blank line.

Issue #271, opened 2020-02-05
License parsing doesn't parse e.g. LGPLv3 correctly


shr-project:
When regenerating recipes for newer eloquent release from 2020-01-24, I've noticed that generated mavlink recipe has license LGPL-2 while the original package.xml has LGPLv3 (https://github.com/mavlink/mavlink-gbp-release/blob/release/eloquent/mavlink/2019.12.30-1/package.xml#L20)

Which for OE triggers this warning:
```
WARNING: mavlink-2019.12.30-1-r0 do_populate_lic: mavlink: No generic license file exists for: LGPL-2 in any provider
```
because LGPL-2 isn't valid SPDX identifier (https://spdx.org/licenses/) nor OE licenses.conf has mapping from LGPL-2 to the valid LGPL-2.0 (http://git.openembedded.org/openembedded-core/tree/meta/conf/licenses.conf#n28)

Mapping LGPLv3 to LGPL-2 is quite wrong and dangerous. I've tried to update the regexps here:
https://github.com/ros-infrastructure/superflore/blob/master/superflore/utils.py#L149
but that's also a bit error prone with the free form of license identifiers people are using in package.xml.

What do you think about using exact mapping for known identifier variants, like licenses.conf in OE does http://git.openembedded.org/openembedded-core/tree/meta/conf/licenses.conf and then mapping them only to valid SPDX identifiers or issuing the warning and passing the value when it's not one of the explicitly supported variants?

That would also cover the issue that LGPL (without any version) is currently being mapped to this LGPL-2 which is very misleading when the actual license in the component is LGPL-3.0.

I have already couple cases in `meta-ros` (mostly from `melodic` release) where this happens:
```
meta-ros1-melodic/recipes-bbappends/navigation/amcl_1.16.3-1.bbappend:# Missing license version in package.xml maps it to invalid LGPL-2
meta-ros1-melodic/recipes-bbappends/navigation/amcl_1.16.3-1.bbappend-LICENSE = "LGPL-2.1+"
meta-ros1-melodic/recipes-bbappends/navigation/amcl_1.16.3-1.bbappend-
--
meta-ros1-melodic/recipes-bbappends/open-karto/open-karto_1.2.0.bbappend:# Missing license version in package.xml maps it to invalid LGPL-2
meta-ros1-melodic/recipes-bbappends/open-karto/open-karto_1.2.0.bbappend-LICENSE = "LGPL-3.0"
--
[...]

PR #279, opened 2020-09-18
Improve license parsing

shr-project:
Fixes https://github.com/ros-infrastructure/superflore/issues/271

Corresponding changes for catkin_pkg:
https://github.com/ros-infrastructure/catkin_pkg/pull/296
and rosdistro:
https://github.com/ros/rosdistro/pull/26601


tests/test_utils.py
```diff
@@ -93,37 +93,37 @@ def test_get_license(self):
...
-        self.assertEqual(ret, 'BSD')
-        ret = get_license('Apache-2')
+        self.assertEqual(ret, 'BSD-3-Clause')
+        ret = get_license('Apache2')
         self.assertEqual(ret, 'Apache-2.0')
         ret = get_license('CreativeCommons-Attribution-NonCommercial-NoDerivatives-4.0')
         self.assertEqual(ret, 'CC-BY-NC-ND-4.0')
         ret = get_license('CC BY-NC-SA 4.0')
         self.assertEqual(ret, 'CC-BY-NC-SA-4.0')
-        ret = get_license('BoostSoftwareLicense Version1.0')
```
allenh1, line 105: While some of my mappings are definitely incorrect (good catch), the inputs to the `get_license` call were usually inspired by actual package xml strings I was failing to parse correctly.

Please let leave the `get_license` as they are (and correct the mappings as appropriate -- again, very good catch).

Feel free to add the changes you made as new tests, though! For something like this, I'm more than happy to do as many as possible.

superflore/utils.py
```diff
@@ -145,67 +144,525 @@ def trim_string(string, length=80):
...
+            'Apachi 2': 'Apache-2.0',
+            'Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)': 'Apache-2.0',  # noqa: E501
+            'Apache v2': 'Apache-2.0',
+            'Apache v2.0': 'Apache-2.0',
+            'Apache2.0': 'Apache-2.0',
+            'APACHE2.0': 'Apache-2.0',
+            'Apache2': 'Apache-2.0',
+            'Apache License, Version 2.0': 'Apache-2.0',
+            'Apache 2': 'Apache-2.0',
+            'Apache 2.0': 'Apache-2.0',
```
allenh1, line 602: @shr-project My conscern is that this will become an unmaintainable, long list of "aliases" for the SPDX identifiers. I really do like the idea of using SPDX, but many of the community packages, unfortunately, fail to be consistent.

This is why I liked the regular expression idea, since it would get "close enough" to correct.

shr-project: There is corresponding PR for catkin_pkg https://github.com/ros-infrastructure/catkin_pkg/pull/296 which shows the license validation warning to the developers and hopefully they will eventually improve the license values in package.xml directly, so this superflore modification would be less and less useful in future until we could drop it completely.

The issue with "close enough" is that in some more "creative" values it can also guess incorrect value and then the bad value from package.xml is still better than wrongly assumed value from regexp.

tests/test_utils.py
shr-project: My mappings are inspired by the values currently included in all package.xml files listed in any distribution in rosdistro and these removed entries don't appear there anymore.

shr-project:
Depends on https://github.com/ros-infrastructure/superflore/pull/281 and https://github.com/ros-infrastructure/superflore/pull/280

superflore/utils.py
```diff
@@ -145,67 +145,554 @@ def trim_string(string, length=80):
...
+                       "YPL-1.1",
+                       "Zed",
+                       "Zend-2.0",
+                       "Zimbra-1.3",
+                       "Zimbra-1.4",
+                       "Zlib",
+                       "zlib-acknowledgement",
+                       "ZPL-1.1",
+                       "ZPL-2.0",
+                       "ZPL-2.1"]
```
allenh1, line 577: :eyes: @shr-project wow, that's got to be all of them now.

allenh1 (approved):
@shr-project this is above and beyond what I was looking for. Thanks a million! :+1: from me.

allenh1:
@shr-project In which order should I merge #279, #280, #281?

Let me know and I'll start hitting merge buttons. :)

shr-project:
> @shr-project In which order should I merge #279, #280, #281?
> 
> Let me know and I'll start hitting merge buttons. :)

#281, #280, #279 :)

herb-kuta-lge (approved):

shr-project:
Please DON'T SQUASH the commits when merging, it will be impossible to review and understand the commits after they are squashed.

herb-kuta-lge (dismissed):

herb-kuta-lge (approved):

herb-kuta-lge:
@allenh1 Even though this PR affects more than the OpenEmbedded portions of the code, since you've already approved it, I'm going to merge it.

PR #280, opened 2020-11-06
Various improvements to support generating recipes for Noetic

shr-project:

PR #281, opened 2020-12-08
Fix parsing of <exports><build_type> with conditionals

shr-project:


shr-project:
Moved from https://github.com/ros-infrastructure/superflore/pull/280 because the 2nd commit might have some impact on the ebuilds for gentoo.

@allenh1 @herb-kuta-lge please review and merge

allenh1 (approved):
:+1:

herb-kuta-lge:
It appears that the two commits were squashed as when the PR was merged.

PR #285, opened 2021-08-06
Noetic support For Gentoo

allenh1:
Signed-off-by: Hunter L. Allen <hunter.allen@ghostrobotics.io>

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -138,12 +138,13 @@ class Ebuild(object):
         # EAPI=<eapi>
         ret = self.get_license_line(distributor, license_text)
         ret += self.get_eapi_line()
+
         if self.python_3 and not self.is_ros2:
             # enable python 2.7 and python 3.5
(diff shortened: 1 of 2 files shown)


tfoote (commented):
Looks reasonable. The chunking is because there's too many to generate at once?

allenh1:
Yeah, that's it exactly. It needs more testing, but seemed to work on my pc when I gave it a go.

Got a bit lazy about pushing updates, and they've stacked up pretty high at this point.

herb-kuta-lge (changes requested):
How about mentioning in the comment message and PR title that the change is "Noetic support for Gentoo" (Noetic is already supported by OE).

robwoolley (approved):
I tested these changes along with the 2 other ebuild PRs.   It successfully generated the recipes but had difficulty with the docker stage.  The error message was "bash: argument list too long".   It may be necessary to reduce the split_limit.   I am approving these changes so that we can establish a new baseline for ebuild and fix any remaining problems.

PR #291, opened 2022-09-28
noetic and python10 support

svavs:
Test done with generation of Humble ebuilds


svavs:
Sorry, my error. Same PR already present.

PR #296, opened 2023-04-10
Update of the manifest generation

tomkimsour:
I updated the dockerfile ebuilds are built on and pushed it on docker hub. The current testing dockerfile is now under tomkimsour/ros_gentoo_base and the dockerfile can be found on ros-overlay repository once the pr I made gets accepted.

I also updated EAPI of the generated ebuild as well as the PYTHON_COMPAT.

### DIFF
--- a/superflore/generators/ebuild/ebuild.py
+++ b/superflore/generators/ebuild/ebuild.py
@@ -50,7 +50,7 @@ class Ebuild(object):
     This is where any necessary variables will be filled.
     """
     def __init__(self):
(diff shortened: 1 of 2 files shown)


robwoolley:
Are these changes still required given the other Gentoo changes that were merged in?

tomkimsour:
I am unsure. It's been a while I've worked on this. I think we can close this PR, if it was indeed necessary and anyone needs it. This can be reopen.

PR #298, opened 2023-05-16
Add "license-org" and "license-text" parameters

JasonShigit:
If "license-org" and license-text" are passing in command line, the default value will be overlayed in generated files.


JasonShigit:
@allenh1 Could you please review and comment?
Test pass with my meta-ros env.

allenh1:
@JasonShigit sorry for the delay! I'm running CI on it now.

allenh1 (dismissed):
LGTM with green CI

JasonShigit:
Hi @allenh1 
Thanks for help. The test failed and i updated one more change for test_ebuild.py. Is there any guidance for running tests locally?

allenh1:
@JasonShigit it seems there are still a few errors to resolve.

The best way to run locally is to install all the requirements, and run pytest.

robwoolley:
Thanks for submitting this.   Can you describe the use case for allowing superflore to generate recipes under a difference license?

Since the recipes are generated based on metadata, I am skeptical whether they *can* be copyrighted in the first place.  Moreover the default copyright used for bitbake recipes is MIT and this patch would change the default to BSD.

If you can describe the reason for this change it will help me better understand why it is needed.   I think before I could merge this in, I would need to run it by the OSRA.

robwoolley:
I raised the questions in our ROS Infra meeting.  The general consensus is that this change should be put on hold for now.  They reached out to the ROS PMC for clarification

PR #299, opened 2023-09-23
Update ebuild generator to support the latest systems

gentoo90:
* fix sporadical empty `*.ebuild` and `metadata.xml` files
* `EAPI6` is dead, so switch to `EAPI8`
* `python` `3.5` and `3.6` are dead, so switch to `3.10` and `3.11`
* fix dependencies of the packages by evaluating conditions
* update syntax of applying patches
* add `buildtool_export` and `build_export` to the dependencies


nuclearsandwich:
@gentoo90 would you be willing to update the GitHub actions-based CI so that we can see test feedback on these changes. If the CI will pass we'll merge these changes.

gentoo90:
@nuclearsandwich, I've updated Ubuntu and Python versions in the CI config. Is there anything else I need to do?

robwoolley:
Thanks for submitting these upgrades for the Gentoo support.   I am not a Gentoo user, myself.  Do you know if these changes are still needed given the other Gentoo fixes from @allenh1 that recently were merged?

PR #301, opened 2024-03-29
Update GitHub Actions workflow

robwoolley:
I saw that you asked for the CI workflow to be updated for #299 

I did a quick run at making the necessary changes.  The commits are clean however since I was working in the GHA workflow editor some commits failed and had to be corrected.  (eg switching from nose to nose3 then to pynose) I am happy to rework the series if you feel it is necessary.

You may see the workflow test runs I did here: https://github.com/robwoolley/superflore/actions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,7 @@ jobs:
         run: |
           sudo rosdep init
           rosdep update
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):
If it's possible to switch to pytest rather than keeping along with nose that's ideal, but if it's complicated and this works it can be punted on. It can also be done as a follow-up PR which is why I'm approving as-is.

Issue #305, opened 2024-11-30
Test source code for PEP8 conformance


robwoolley:
The CI tests currently fail to test for PEP8 conformance:

```
======================================================================
FAIL: Test source code for PEP8 conformance
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.15/x64/lib/python3.10/site-packages/nose/case.py", line 170, in runTest
    self.test(*self.arg)
  File "/home/runner/work/superflore/superflore/tests/test_code_format.py", line 54, in test_pep8
    assert p.returncode == 0, \
AssertionError: Command '/opt/hostedtoolcache/Python/3.10.15/x64/bin/python -m pycodestyle /home/runner/work/superflore/superflore/tests/../superflore --count' returned non-zero exit code '1'
```

Running the test manually shows:
```
(venv) rosuser@da759d65bbf0:~$ python -m pycodestyle superflore/superflore
superflore/superflore/generators/ebuild/gen_packages.py:159:80: E501 line too long (88 > 79 characters)
superflore/superflore/generators/ebuild/gen_packages.py:177:80: E501 line too long (99 > 79 characters)
superflore/superflore/generators/ebuild/gen_packages.py:212:80: E501 line too long (88 > 79 characters)
superflore/superflore/generators/ebuild/overlay_instance.py:99:80: E501 line too long (80 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:149:80: E501 line too long (96 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:207:80: E501 line too long (107 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:213:80: E501 line too long (86 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:222:80: E501 line too long (88 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:255:80: E501 line too long (88 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:258:80: E501 line too long (100 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:278:80: E501 line too long (84 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:280:80: E501 line too long (115 > 79 characters)
superflore/superflore/generators/buildstream/bst_element.py:315:80: E501 line too long (81 > 79 characters)
[...]

PR #308, opened 2024-12-11
Fix code style guide problems for PEP8 and Flake8

robwoolley:
The CI system picked up non-conformance issues with recent changes to the ebuild generator.

Fix lines that exceed 79 characters and remove an used chunk_count variable.


robwoolley:
This resolves issues raised by #304 and #305

nuclearsandwich (approved):

PR #315, opened 2025-07-03
Replace pkg_resources usage with importlib.metadata.

### LINKED ISSUE #314: Replace pkg_resources usage with importlib
"The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81."

The bloom tool also has a need to replace pkg_resources with importlib:  https://github.com/ros-infrastructure/bloom/pull/747

nuclearsandwich:
pkg_resources is only used to fetch and provide the current superflore version.

In fact, there's very little need to parse this version at all (which would also require adding the `packaging` module as a dependency. Instead of parsing the version for the test, I used a cursed regexp that will test the value is greater than the current release when present.

Resolves #314.

### DIFF
--- a/superflore/__init__.py
+++ b/superflore/__init__.py
@@ -1,11 +1,6 @@
-try:
-    import pkg_resources
-    try:
-        __version__ = pkg_resources.require('superflore')[0].version
-    except pkg_resources.DistributionNotFound:
-        __version__ = 'unset'
-except (ImportError, OSError):
-    __version__ = 'unset'
+from superflore.version import VERSION
+
+__version__ = VERSION
 
 from .repo_instance import RepoInstance
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -17,8 +17,7 @@ import string
 import sys
 import time
 
-from pkg_resources import parse_version
-from superflore import __version__
+from superflore.version import VERSION
 from superflore.exceptions import UnknownPlatform
 from superflore.exceptions import UnresolvedDependency
 from superflore.TempfileManager import TempfileManager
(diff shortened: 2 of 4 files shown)


superflore/__init__.py
```diff
@@ -1,11 +1,6 @@
-try:
-    import pkg_resources
-    try:
-        __version__ = pkg_resources.require('superflore')[0].version
-    except pkg_resources.DistributionNotFound:
-        __version__ = 'unset'
-except (ImportError, OSError):
-    __version__ = 'unset'
+from superflore.version import VERSION
```
nuclearsandwich, line 1: This convoluted indirection was an attempt to work around local import trouble with the RepoInstance. It may be we don't need the superflore.version module if the tests are better behaved in general than they are in my local workspace.

robwoolley:
Thanks @nuclearsandwich for tackling this.

It seems that the new lines have a new Flake8 error:
```
superflore/utils.py:29:1: I100 Import statements are in the wrong order. 'from superflore.exceptions import UnknownPlatform' should be before 'from superflore.version import VERSION'
superflore/utils.py:29:1: I202 Additional newline in a group of imports. 'from superflore.exceptions import UnknownPlatform' is identified as Third Party and 'from superflore.version import VERSION' is identified as Third Party.
superflore/version.py:11:1: W391 blank line at end of file
```

It seemed to be happy with this modification:
```
git diff superflore/utils.py
diff --git a/superflore/utils.py b/superflore/utils.py
index b549beb..b19500d 100644
--- a/superflore/utils.py
+++ b/superflore/utils.py
@@ -24,10 +24,9 @@ import time
 from typing import Dict
 import urllib.request

-from superflore.version import VERSION
-
 from superflore.exceptions import UnknownPlatform
 from superflore.rosdep_support import get_cached_index, resolve_rosdep_key
+from superflore.version import VERSION
 from termcolor import colored
```

robwoolley (approved):

tests/test_utils.py
```diff
@@ -261,6 +260,7 @@ def callback_retries(expected_num_retries):
 
     def test_get_superflore_version(self):
         """Test get SuperFlore version"""
-        if __version__ != 'unset':
-            self.assertGreaterEqual(parse_version(get_superflore_version()),
-                                    parse_version('0.2.1'))
+        if VERSION != 'unset':
+            self.assertRegex(get_superflore_version(),
+                             r"^(\d{2}|[1-9]|0\.(\d{2}|[4-9]|3\.(\d{2}|[4-9]|3)))",
```
nuclearsandwich, line 265: Surprised (and a little disappointed) that my use of a cursed regexp to do the version parsing went unremarked.

PR #317, opened 2025-07-06
oelint: drop the space at the first line in a multiline assignment

vermaete:
This fixes this issue with the oelint-adv tool:
  info:oelint.vars.notneededspace:Space at the beginning of the var is not needed [branch:true]


vermaete:
@robwoolley Do you need something to be done by me before merging?

robwoolley (approved):
This change looks good.

PR #319, opened 2025-07-06
Remove the 'S = ' in the recipes.

vermaete:
This is not longer needed in Yocto recipes.
And must be removed for the Yocto release Whinlatter and master

This fixes this error:

  ERROR: Recipes that set S = "${WORKDIR}/git" or S = "${UNPACKDIR}/git" should remove that assignment, as S set by bitbake.conf in oe-core now works.

See:
  https://github.com/openembedded/openembedded-core/commit/46480a5e66747a673041fe4452a0ab14a1736d5e


robwoolley:
Thanks for raising this issue, @vermaete.  I have submitting this pull request that allows us to optionally specify the Yocto release when using superflore: https://github.com/ros-infrastructure/superflore/pull/321

It allows us to conditionally include the line for older Yocto releases.

PR #324, opened 2025-08-12
bitbake: Remove source directory variable in Whinlatter

rcwoolley:
Styhead requires that S = "${WORKDIR}" be replaced with S = "${UNPACKDIR}", however S = "${WORKDIR}/git" is still supported.

Whinlatter no longer supports setting S = "${WORKDIR}/git".

See the Yocto Project release notes for details.

### DIFF
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -512,8 +512,9 @@ class yoctoRecipe(object):
         ret += 'SRCREV = "' + self.srcrev + '"\n'
         if (self.release):
             if Version(self._get_yocto_version(self.release)) < \
-               Version(yocto_releases['styhead']):
-                ret += 'S = "${WORKDIR}/git"\n\n'
+               Version(yocto_releases['whinlatter']):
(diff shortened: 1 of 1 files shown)


robwoolley (approved):
Styhead

	https://docs.yoctoproject.org/next/migration-guides/migration-5.1.html#workdir-changes
	S = "${WORKDIR}" becomes S = "${UNPACKDIR}"

	Note: "S = "${WORKDIR}/git" and also deeper paths continue to work as expected without changes."

Whinlatter

	https://docs.yoctoproject.org/next/migration-guides/migration-5.3.html#workdir-changes
	
	"S = ${WORKDIR}/something no longer supported"

	" S setting in recipes can and should be removed."

	when "git" as the source checkout location is hardcoded "in SRC_URI, replace with BB_GIT_DEFAULT_DESTSUFFIX, otherwise replace with BP."


How to make those adjustments without tedious manual editing
	https://docs.yoctoproject.org/next/migration-guides/migration-5.3.html#how-to-make-those-adjustments-without-tedious-manual-editing

PR #325, opened 2025-09-02
Allow running superflore with uncached rosdistro

harrysarson:
This commits adds a new command line flag to force superflore (via the rosdistro python package) to regenerate the distribution information.

This is useful when running superflore with $ROSDISTRO_INDEX_URL set to get deterministic codegen. The rosdistro caches are updated to latest and not versioned, see https://github.com/ros/rosdistro/issues/47469.

### DIFF
--- a/superflore/generators/bitbake/run.py
+++ b/superflore/generators/bitbake/run.py
@@ -15,7 +15,6 @@
 import os
 import sys
 
-from rosinstall_generator.distro import get_distro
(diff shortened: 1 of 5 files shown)


harrysarson:
I am not very familiar with the bitbake/ebuild parts of superflore. Are the CI tests sufficient or are there tests I can run locally to validate these changes?

PR #327, opened 2026-07-24
SPDX: use 'AND' instead of '&'

vermaete:
The SPDX specification no longer accepts '&' as a license operator. Replace it with 'AND' to comply with the SPDX rules used by Yocto.

@robwoolley I have tested it with a local script.  Not with the 'Actions' of Github.  But we do need it at master.  Could you redo the latest sync of master/rolling so I can verify it and finish the recipes in `meta-ros2` and `meta-ros-common`?

### DIFF
--- a/superflore/generators/bitbake/yocto_recipe.py
+++ b/superflore/generators/bitbake/yocto_recipe.py
@@ -56,6 +56,8 @@ yocto_releases = {
     'walnascar': '5.2',
     'whinlatter': '5.3',
     'wrynose': '6.0',
+    'wrynose': '6.0',
+    'blacksail': '6.1',
 }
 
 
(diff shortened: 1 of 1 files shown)


robwoolley:
Thanks @vermaete.  I can't merge this as it is as it would break support for Scarthgap and Wrynose.  There is already an example of making changes based on the target Yocto release.  

See here for details: https://github.com/ros-infrastructure/superflore/blob/master/superflore/generators/bitbake/yocto_recipe.py#L513

Would it be possible to rework this PR to make it only applicable to Blacksail and later releases?

vermaete:
@robwoolley I have added the check on the Yocto version.

robwoolley:
What do you think of adding the whitespace to _concat (eg. " AND " and " & ") to avoid having to use the formatted string?

eg.
oe_lic = f' {_concat} '.join([get_license(lic) for lic in self.license])

becomes:
oe_lic = _concat.join([get_license(lic) for lic in self.license])

robwoolley:
This change looks correct but failed in testing because of how scripts/ros-generate-recipes.sh works.  I need to fix the script to validate this fix.

See https://github.com/ros/meta-ros/issues/1814

robwoolley (commented):
This is great.  Thank you.  I have run CI builds to confirm that this is working.

superflore/generators/bitbake/yocto_recipe.py
robwoolley, line 1: Approved

robwoolley (approved):
PEP8 is finally satisfied with the line lengths.

PR #329, opened 2026-08-20
Superflore Tooling Modernization

robwoolley:
I used Claude to assist with modernizing the tooling for Superflore.  The details are listed below.  The commits have been manually reviewed for accuracy.

This pull request is being shared to gather feedback from downstream users of superflore and to perform additional testing on GitHub.

---

Goal: modernize packaging, linting, and testing tooling with no change to
runtime behavior. Milestones are ordered must-do → recommended → nice-to-have
(see the accompanying recommendations for the full rationale and tradeoffs).
Flag-only items (`stdeb.cfg`, `setup_superflore.sh`) are intentionally
excluded — they need an owner decision before any milestone touches them.

Each milestone should land as its own PR so CI validates it independently.

---

### M1 — Migrate packaging to `pyproject.toml`

**Depends on:** nothing (do first).

**Tasks:**
- Add `pyproject.toml` with `[project]` metadata (name, description, authors,
  license, classifiers, `python_requires`, `console_scripts` entry points)
  ported from [setup.py](setup.py).
- Adopt `setuptools-scm` for versioning, replacing the custom
  `append_local_version_label()` git-describe logic. Configure via
  `[tool.setuptools_scm]`; confirm the resulting version string is still
  PEP 440-compliant and acceptable to anything that parses it downstream
  (release scripts, changelog tooling).
- Split dependencies:
  - Runtime deps (`xmltodict`, `termcolor`, `gitpython`, `requests`,
    `docker`, `pyyaml`, `pygithub`, `rosinstall_generator`, `rosdistro`,
    `rosdep`, `catkin_pkg`, `rospkg`, `setuptools`) → `[project.dependencies]`.
  - Dev/lint/test tools (`pynose`, `flake8`, `flake8-import-order`, `pep8`)
    → a `dev`/`test` group under `[project.optional-dependencies]` or
    `[dependency-groups]`. These get replaced in M3/M2 but should move as-is
    here so this milestone stays packaging-only.
- Remove `setup.py` once `pyproject.toml` fully covers it, unless
  setuptools-scm or the build backend needs a minimal shim — confirm with a
  clean `pip install .` and `python -m build`.
- Decide whether `requirements.txt` is retired (replaced by
  `pip install -e ".[dev]"`) or kept as a pinned/lock-style install path.

**Acceptance criteria:**
- `pip install .` and `pip install -e ".[dev]"` both work from a clean venv.
- `python -m build` produces a valid sdist + wheel with correct version
[...]

PR #330, opened 2026-09-08
Modernize Tooling V2 - Part 1/2

robwoolley:
I have split the original PR: https://github.com/ros-infrastructure/superflore/pull/329 into 2 parts

The first part handles:

- M1 — Migrate packaging to pyproject.toml
- M2 — Migrate test suite from nose to pytest

The second part will include things related to formatting and CI:

- M3 — Consolidate linting/formatting onto Ruff
- M4 — Add pre-commit configuration
- M5 — Update CI workflow


tests/test_docker.py
```diff
@@ -16,6 +16,9 @@
 from superflore.docker import NoDockerfileSupplied
 import unittest
 
+@unittest.skip(
```
tfoote, line 19: You've switched over to pytest so you can use `@pytest.mark.nvidia` (where nvidia is the tag) like we do in rocker https://github.com/osrf/rocker/blob/00bf38f685ffab5d5a250e484bd13809161cb08a/test/test_nvidia.py#L37

Then you can  use -m "not docker" like we skip the nvidia tags here: https://github.com/osrf/rocker/blob/00bf38f685ffab5d5a250e484bd13809161cb08a/.github/workflows/basic-ci.yaml#L27 

Though we do test docker things in CI on rocker so I'd double check your assumption that you can't do anything with docker. And this is another reason to use the flags so that you can choose to test with our without the docker tests depending on your platform without changing the source.

tfoote (approved):
One small tweak for the skips requested. I'll approve but strongly suggest switching to the mark and exclude versus just straight skipping.

tests/test_docker.py
robwoolley: Thanks, @tfoote!

I really appreciate that feedback.  I will create a new issue to follow up on that.

PR #332, opened 2026-09-10
Modernize Tooling V2 - Part 2/2

robwoolley:
This is a continuation of  part 1 (https://github.com/ros-infrastructure/superflore/pull/330) which came from Superflore Tooling Modernization (https://github.com/ros-infrastructure/superflore/pull/329)

M3 — Consolidate linting/formatting onto Ruff
M4 — Add pre-commit configuration
M5 — Update CI workflow
M6 — Introduce type checking (incremental)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.16.3
```
claraberendsen: I would suggest pinning to a SHA here to avoid security vulnerabilities with tag updates. 

The usual pre-commit update workflow should work but you need to pass the `--freeze` option. 

```
pre-commit autoupdate --freeze
```

